PR #3, opened 2015-09-25
entry: Add FindRelative function

andaru:
FindRelative finds the Entry relative to the entry, given a relative
path. nil is returned if no matching entry was found.


andaru:
Dropping this in favor of Paul's #8

PR #8, opened 2015-10-09
Add relative traversal to Find based on andaru's FindRelative.

pborman:
Andrew,

I believe this is your FindRelative pulled into Find.


andaru:
:shipit:

LGTM
Thanks!

I'll add some more tests next week when I get back from my trip if you like.

Issue #7, opened 2015-10-09
prefix handling with grouping/uses is ineffective


gopenconfig:
For a `yang.Entry`, `Prefix` is only set when the current node has a `prefix` statement. This is only the case for the entry corresponding a `yang.Module`.

When a `grouping` statement is "entry-ised", the entry corresponding the `grouping` has a nil `Parent`. This means that when the `grouping` is placed in a `uses` (i.e., in some other module), we will never be able to ascend the `Parent` entries to determine the current prefix.

One solution would be to simply set the `Prefix` for each entry.

andaru:
A test showing a related set of issues (with a simple 'offending' GetPrefix implementation) can be seen in a related branch in a dev repo I'm working on:

https://github.com/andaru/goyang/tree/andaru/test-getprefix

This commit in particular:

https://github.com/andaru/goyang/commit/2de1f1e92cd8968c1fb7cb9b49a4bc77c510252a

This GetPrefix impl wouldn't be necessary if the Prefix was set for each entry during ToEntry.

robshakir:
@andaru -- I think that you closed this one out too, AFAICS? I also fixed `namespace` on this.

andaru:
Thanks, Rob; after your fixes recently this is all done.

Issue #5, opened 2015-10-01
prefix handling is incomplete


pborman:
goyang tends to ignore prefixes.  In openconfig we do not expect to see conflicts on names sans prefixes, but for general use, goyang must support prefixes better.

andaru:
In our present usages, one area we've run into a concern with not having prefixes is for type names - since types are often pulled in from other modules (e.g., common prefix `inet`), so we describe the type name as `inet:as-number` in one place, and `as-number` in another, because we receive only `as-number` back from goyang.

It thus seems preferable for users who do care about prefixes to get back the fully qualified type name (i.e., that with the _source_ module prefix) from things like `(*yang.Entry).Type.Name`.

r-cc-c:
Just FYI. I got this error when parsing openconfig-bgp: "../policy/openconfig-policy-types.yang:123:7: unknown prefix: yang for type hex-string". pyang didn't complain about this.

pborman:
I believe pull request 60 may resolve this.

r-cc-c:
Thanks pborman. FYI,  I just tried the latest (with #60 and #61). Same error appear.

pborman:
Can you provide instructions on how to reproduce this, including where all the .yang files have come from and what version?  It would be great if you could just say

mkdir foo
cd foo
git clone ....
git clone ....
git cone https://github.com/openconfig/goyang
cd goyang
go build
./goyang ...

So I can reproduce exactly what you are seeing.

r-cc-c:
Thanks Again! I discovered on my script the policy-types.yang file appear twice. It was my bad. 
Thanks for your help!

rski:
Looks like leafs (and maybe leaft-lists and uses leaf-lists) don't have prefixes set, because ToEntry() doesn't set a prefix in these cases.

Here is a snipper that shows that:

```git clone https://github.com/openconfig/public /tmp/public```

```go
package main

import (
	"fmt"
	"os"
	"path/filepath"
	"strings"

	"github.com/openconfig/goyang/pkg/yang"
)

func main() {
	yangDir := "/tmp/public/release/models"

	ms := yang.NewModules()
	filepath.Walk(yangDir, func(path string, info os.FileInfo, err error) error {
		if !strings.HasSuffix(path, ".yang") {
			return nil
		}
		return ms.Read(path)
	})

	for _, k := range ms.Modules["openconfig-isis"].Grouping {
		if k.Name == "isis-overload-bit-reset-config" {
[...]

PR #11, opened 2015-11-11
Set every Entry's Prefix 

andaru:
Change `Entry.Prefix` from a string to a `*Value`, set it based on the current `Module`s `Prefix` if it is not explicitly set.

Add a test to confirm we get the correct prefix for a `uses` case across YANG modules.

This is designed to address issue #7, the first issue related to a parent issue (#5)

### DIFF
--- a/pkg/yang/yang.go
+++ b/pkg/yang/yang.go
@@ -179,14 +179,23 @@ func (s *Module) FullName() string {
 // GetPrefix returns the proper prefix of m.  Useful when looking up types
 // in modules found by FindModuleByPrefix.
 func (m *Module) GetPrefix() string {
-	if m == nil {
+	pfx := m.getPrefix()
+	if pfx == nil {
 		// This case can be true during testing.
 		return ""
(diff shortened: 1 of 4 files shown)


pkg/yang/yang.go
```diff
@@ -179,17 +179,26 @@ func (s *Module) FullName() string {
...
+}
+
+func (m *Module) getPrefix() *Value {
+	if m == nil {
+		return nil
+	} else if m.Prefix != nil {
+		return m.Prefix
+	} else if m.BelongsTo != nil {
+		return m.BelongsTo.Prefix
 	}
```
pborman: This should be written as a switch statement:

switch {
case m == nil:
    return nil
case m.Prefix != nil:
    return m.Prefix
case m.BelongsTo != nil:
    return m.BelongsTo.Prefix
default:
    return nil
}

andaru: Done in the following commit

PR #15, opened 2016-02-19
yang: Add a Namespace() function to Entry

andaru:
The `Namespace()` function on Entry returns the namespace value for an entry
as described by RFC6020 section 7.12. A grouping's namespace is defined
by the namespace of the module using them, rather than the module
defining them.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -726,6 +726,30 @@ func (e *Entry) Path() string {
 	return e.Parent.Path() + "/" + e.Name
 }
 
+// Namespace returns the YANG/XML namespace Value for e as mounted in the Entry
+// tree (e.g., as placed by grouping statements).
+//
+// Per RFC6020 section 7.12, the namespace on elements in the tree due to a
+// "uses" statement is that of the where the uses statement occurs, i.e., the
+// user, rather than creator (grouping) of those elements, so we follow the
+// usage (Entry) tree up to the parent before obtaining the (then adjacent) root
+// node for its namespace Value.
+func (e *Entry) Namespace() *Value {
+	// Make e the root parent entry
+	for ; e.Parent != nil; e = e.Parent {
+	}
+
+	// Return the namespace of a valid root parent entry
+	if e != nil && e.Node != nil {
+		if root := RootNode(e.Node); root != nil {
+			return root.Namespace
+		}
+	}
+
+	// Otherwise return an empty namespace Value (rather than nil)
+	return new(Value)
+}
+
 // dup makes a deep duplicate of e.
 func (e *Entry) dup() *Entry {
 	// Warning: if we add any elements to Entry that should not be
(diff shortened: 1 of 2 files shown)


pkg/yang/entry.go
```diff
@@ -726,6 +726,30 @@ func (e *Entry) Path() string {
...
+// tree (e.g., as placed by grouping statements).
+//
+// Per RFC6020 section 7.12, the namespace on elements in the tree due to a
+// "uses" statement is that of the where the uses statement occurs, i.e., the
+// user, rather than creator (grouping) of those elements, so we follow the
+// usage (Entry) tree up to the parent before obtaining the (then adjacent) root
+// node for its namespace Value.
+func (e *Entry) Namespace() *Value {
+	// Make e the root parent entry
+	for ; e.Parent != nil; e = e.Parent {
```
pborman: So what if there is a uses inside of a uses?  Is the namespace the inner or outer namespace?  I realize this code asserts the outer, but I am wondering if you thought about that.

pkg/yang/entry.go
```diff
@@ -726,6 +726,30 @@ func (e *Entry) Path() string {
...
+
+	// Return the namespace of a valid root parent entry
+	if e != nil && e.Node != nil {
+		if root := RootNode(e.Node); root != nil {
+			return root.Namespace
+		}
+	}
+
+	// Otherwise return an empty namespace Value (rather than nil)
+	return new(Value)
```
pborman: I prefer &Value{}

pkg/yang/entry.go
andaru: I think the following text found in [this note](https://tools.ietf.org/html/rfc6020#section-6.2.1) is the relevant section from RFC6020.

Unfortunately, I'm not certain whether "parent" in the below means the immediate or the ultimate parent node.

```
   o  All leafs, leaf-lists, lists, containers, choices, rpcs,
      notifications, and anyxmls defined (directly or through a uses
      statement) within a parent node or at the top level of the module
      or its submodules share the same identifier namespace.  This
      namespace is scoped to the parent node or module, unless the
      parent node is a case node.  In that case, the namespace is scoped
      to the closest ancestor node that is not a case or choice node.
```

What do you think?

pkg/yang/entry.go
andaru: Will do in following update to the PR. Thanks.

pkg/yang/entry.go
pborman: Okay, this appears to be the correct logic.  In 7.12 it states:

>    The identifiers defined in the grouping are not bound to a namespace
>    until the contents of the grouping are added to the schema tree via a
>    "uses" statement that does not appear inside a "grouping" statement,
>    at which point they are bound to the namespace of the current module.

So a uses inside a grouping does not dictate the name space, only the uses in the main module do.  Submodules share the same name space as the main module, so even if they use a uses as the top level, it is still in the same name space as the main module.  Importing modules, on the other hand, only brings in definitions, it does not change the schema tree.  A uses inside of a group doesn't immediately change the schema tree, either.  The schema tree is only changed once the main module (or submodule) uses a uses outside a grouping.

pkg/yang/entry.go
pborman: Please do.

PR #16, opened 2016-08-19
Add the Extra map to an Entry.

pborman:
The Extra map is a map of all fields that have no specific processing.
The map is keyed on the field name (such as "revision").  The value
is a slice of interface{} (in case the statement is seen more than
once).  It is the responsibility of client code to map the elements
of the slice to their approrpiate type (e.g., *yang.Revision).

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -89,6 +89,9 @@ type Entry struct {
 	RPC *RPCEntry // set if we are an RPC
 
 	Augments []*Entry // Augments associated with this entry
+
+	// Extra maps all the unsupported fields to thier values
+	Extra map[string][]interface{}
 }
 
 // An RPCEntry contains information related to an RPC Node.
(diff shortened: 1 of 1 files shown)


pkg/yang/entry.go
```diff
@@ -89,6 +89,9 @@ type Entry struct {
 	RPC *RPCEntry // set if we are an RPC
 
 	Augments []*Entry // Augments associated with this entry
+
+	// Extra maps all the unsupported fields to their values
+	Extra map[string][]interface{}
```
marcushines: can't we do something alittle better than interface{} like at least type this to something reasonable or do we really not know what this can be?

pborman: It can be any of over two dozen types.  This basically lets you extend the compiler in client code.  The compiler gets an interface{} and type asserts it to the type that must be produced by that particular field.

PR #23, opened 2016-10-03
Add support for the Identity statement

robshakir:
Hi @pborman,

As discussed, this adds support for the `identity` statement to goyang. There are a couple of things that I'd appreciate your thoughts on:

* There seemed to be a conflict with `Base` in the `Type` struct, so I created separate bases for types (which the original code seemed to use) and `IdentityBase` specifically for the base statement of a YANG `identityref`.
* I added some basic test cases, I'll expand on these as I add some more functionality (building a resolved tree of identities). However, I split the commit up so that the PR wasn't too large.

PTAL :-) Comments on things that don't look good Go best practice are much appreciated!

r.

```
  * (M) doc.go:
    - fix minor typo.
  * (M) entry.go:
    - add support for an Entry to have a defined Identities
      struct member which stores the identities defined within
      that context.
    - add a new entry within the map of statements to indicate
      where a statement is supported, but doesn't require specific
      handling.
  * (M) types.go:
    - fix minor typo.
    - add support for the identityref's base statement and copy
      the value to the IdentityBase member of the entry struct.
    - ensure that an identityref without a base is considered
      erroneous.
  * (M) types_test.go:
    - add test for identityref without a base statement.
  * (M) yang.go:
    - add entries to the YANG structures storing the base of
      an identity, and an getter function for the identities of a
      module.
  * (A) identity_test.go
    - add tests for identities - created separate file based on
      need for resolution code.
```

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -88,6 +88,10 @@ type Entry struct {
 
 	RPC *RPCEntry // set if we are an RPC
 
+	// Identities that are defined in this context, this is set if the Entry
+	// is a module only.
+	Identities []*Identity
+
 	Augments []*Entry // Augments associated with this entry
 
 	// Extra maps all the unsupported fields to their values
@@ -544,6 +548,10 @@ func ToEntry(n Node) (e *Entry) {
 				e.RPC.Output.Name = "output"
 				e.RPC.Output.Kind = OutputEntry
 			}
+		case "identity":
+			if i := fv.Interface().([]*Identity); i != nil {
+				e.Identities = i
+			}
 		case "uses":
 			for _, a := range fv.Interface().([]*Uses) {
 				e.merge(nil, ToEntry(a))
@@ -553,6 +561,10 @@ func ToEntry(n Node) (e *Entry) {
 			// BUG(borman): I think a deviate statement might trigger this.
 			e.addError(fmt.Errorf("%s: unexpected type in %s:%s", Source(n), n.Kind(), n.NName()))
 
+		// Keywords that do not need to be handled as an Entry as they are added
+		// to other dictionaries.
+		case "typedef":
+			continue
 		// TODO(borman): unimplemented keywords
 		case "belongs-to",
 			"contact",
@@ -560,7 +572,6 @@ func ToEntry(n Node) (e *Entry) {
 			"deviation",
 			"extension",
 			"feature",
-			"identity",
 			"if-feature",
 			"mandatory",
 			"max-elements",
@@ -573,7 +584,6 @@ func ToEntry(n Node) (e *Entry) {
 			"reference",
 			"revision",
 			"status",
-			"typedef",
 			"unique",
 			"when",
 			"yang-version":
--- a/pkg/yang/identity_test.go
+++ b/pkg/yang/identity_test.go
@@ -1,0 +1,141 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package yang
+
+import "testing"
+
+type modIn struct {
+	name    string
+	content string
+}
+
+type modOut struct {
+	name       string
+	identities []*idOut
+}
+
+type idOut struct {
+	idName   string
+	baseName string
+}
+
+type testCase struct {
+	name string
+	in   []*modIn
+	out  *modOut
+	err  string
+}
+
+// A set of test modules to parse as part of the test case.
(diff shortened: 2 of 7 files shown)


robshakir:
I actually made some slight changes to this code in https://github.com/robshakir/goyang/tree/robjs-identity-resolved - but this functionality (resolving the identity's base statement to a pointer to the Identity struct, and doing this hierarchically) builds on this commit. I'll wait for approval on this one before sending a PR for the other branch.

Thanks!
r.

pkg/yang/entry.go
```diff
@@ -544,6 +548,10 @@ func ToEntry(n Node) (e *Entry) {
 				e.RPC.Output.Name = "output"
 				e.RPC.Output.Kind = OutputEntry
 			}
+		case "identity":
+			if i := fv.Interface().([]*Identity); i != nil {
+				e.Identities = i
+			}
```
pborman: You don't need to append them?

pkg/yang/identity_test.go
```diff
@@ -0,0 +1,141 @@
...
+	identities []*idOut
+}
+
+type idOut struct {
+	idName   string
+	baseName string
+}
+
+type testCase struct {
+	name string
```
pborman: You should set this in your test cases.

pkg/yang/identity_test.go
```diff
@@ -0,0 +1,141 @@
...
+	in   []*modIn
+	out  *modOut
```
pborman: this can just be:

out map[string]string  // maps identity name to base name

then later, entries like:

out: map[string]string {
    "TEST_ID": "",
    "TEST_CHILD": "TEST_ID",
}

pkg/yang/identity_test.go
```diff
@@ -0,0 +1,141 @@
...
+					baseName: "TEST_ID",
+				},
+			},
+		},
+		err: "could not resolve identities",
+	},
+}
+
+func TestIdentityExtract(t *testing.T) {
+	for _, testCase := range testCases {
```
pborman: It is more typical to see "tt" here, rather than testCase.

pkg/yang/identity_test.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+type idOut struct {
+	idName   string
+	baseName string
+}
+
+type testCase struct {
+	name string
+	in   []*modIn
```
pborman: Don't bother making this a reference, just make it []modIn

pkg/yang/identity_test.go
```diff
@@ -0,0 +1,141 @@
...
+			_ = ms.Parse(mod.content, mod.name)
+		}
+		parsedMod, err := ms.GetModule(testCase.out.name)
+
+		if err != nil {
+			t.Errorf("could not parse module: %s", testCase.out.name)
+			continue
+		}
+
+		identityNames := make(map[string]*Identity)
```
pborman: I don't think  you need this map once you have out being a map:

for _, id := range parseMod.Identities {
    base, ok := tt.out[id.Name]
    delete(tt.out, id.Name)
    switch {
    case !ok:
        t.Errorf("%s: identity %q missing", tt.name, id.Name)
    case base != id.Base.Name:
        t.Errorf("%s: identity %q got base %q, want %q", tt.name, id.Name, id.Base.Name)
    }
}
for id := range tt.out {
    t.Errorf("%s: did not get expected identity %q", tt.name, id)
}

pkg/yang/types_builtin.go
```diff
@@ -246,9 +246,13 @@ func (e *EnumType) ValueMap() map[int64]string {
...
-	Name             string
-	Kind             TypeKind    // Ynone if not a base type
-	Base             *Type       `json:"-"` // dervied from
+	Name string
+	Kind TypeKind // Ynone if not a base type
+	// Base represents the base Type if this is a derived type
+	Base *Type `json:"-"` // dervied from
+	// IdentityBase represents the "base" statement of an Identityref type.
+	// It is distinct from the Base specified above.
+	IdentityBase     *Value
```
pborman: Try to keep these as line comments, rather than block comments:

Base *Type `json:"-"` // base type for derived types
IdentityBase *Value  // identity base statement for identityrefs

robshakir:
Thanks Paul! Review much appreciated.

I refactored the test case file in the subsequent pull request, so I'll go through the comments there, and close this PR since its subsumed by the other one anyway, I'll make the changes noted above in that branch. LMK if this doesn't work.

pkg/yang/types_builtin.go
robshakir: Done, thanks.

pkg/yang/identity_test.go
robshakir: Per the comments from @andaru via email, I modified all references within the `identity_test` to make them directly include the relevant entity rather than reference it.

pkg/yang/identity_test.go
robshakir: Thanks, renamed.

pkg/yang/identity_test.go
robshakir: Thanks. I refactored how this was set up in the other PR, I'll take a look there and see if I can make the test cases there correspond to this (the check becomes more than just the children there).

pkg/yang/identity_test.go
robshakir: Thanks. Done.

PR #25, opened 2016-10-05
Add support for recursing into path directories.

robshakir:
For OpenConfig models, each module sits within a subdirectory of the main "modules" directory -- currently, this means that goyang needs `--path=$OCDIR/mod1,$OCDIR/mod2` to be specified. This patch adds support for recursing into the directories that are specified in --path, such that it is possible to simply use `--path=$OCDIR`.

```
 * (M) yang.go
   - Where a user specifies a directory as a new path (--path) then
     recurse into that directory to find all subdirectories and
     include these in the path. This limits the command-line length
     where particular directories of YANG module files exist.
```


pborman:
Did you try just appending ... to the path?  e.g.,  some/path/... should include some/path and all of its subdirectories.

robshakir:
I didn't, and that does work without this. I'll close this PR, thanks!

PR #26, opened 2016-10-06
Add support for resolving the values of an Identity

robshakir:
This PR covers functionality built on top of #23. In addition to determining the Identities that exist, this functionality resolves identities to their respective bases (including if there are multiple levels of inheritance). In addition, add support for `identityref` leaves to have `IdentityBase` populated with the relevant identity - the types that are valid for that `identityref` to refer to are then populated in `IdentityBase.Values`.

Minor other fixes - adding test case coverage, and supporting `identityref` within a `typedef` also included.

### DIFF
--- a/pkg/yang/identity.go
+++ b/pkg/yang/identity.go
@@ -22,29 +22,38 @@ import (
 // This file implements data structures and functions that relate to the
 // identity type.
 
+// identityDictionary stores a global set of identities that have been resolved
+// to be identified by their module and name.
 type identityDictionary struct {
 	mu   sync.Mutex
-	dict map[string]*resolvedIdentity
+	dict map[string]resolvedIdentity
 }
 
 // Global dictionary of resolved identities
+var identities = identityDictionary{dict: map[string]resolvedIdentity{}}
 
-var identityDict = identityDictionary{dict: map[string]*resolvedIdentity{}}
-
+// resolvedIdentity is an Identity that has been disambiguated.
 type resolvedIdentity struct {
 	Module   *Module
 	Identity *Identity
-	//Children []*resolvedIdentity
 }
 
-func makeResolvedIdentity(m *Module, i *Identity) (string, *resolvedIdentity) {
+// isEmpty determines whether the resolvedIdentity struct value was defined.
+func (r resolvedIdentity) isEmpty() bool {
+	if r.Module == nil && r.Identity == nil {
+		return true
+	}
+	return false
+}
+
+// newResolvedIdentity creates a resolved identity from an identity and its
+// associated value.
+func newResolvedIdentity(m *Module, i *Identity) (string, *resolvedIdentity) {
 	r := &resolvedIdentity{
 		Module:   m,
 		Identity: i,
 	}
-	prefix := RootNode(i).GetPrefix()
-	keyName := fmt.Sprintf("%s:%s", prefix, i.Name)
-	return keyName, r
+	return i.PrefixedName(), r
 }
 
 func appendIfNotIn(ids []*Identity, chk *Identity) []*Identity {
@@ -56,22 +65,21 @@ func appendIfNotIn(ids []*Identity, chk *Identity) []*Identity {
 	return append(ids, chk)
 }
 
-func learnChildren(r *Identity, ids []*Identity) []*Identity {
-	// Learn this child
+// addChildren recursively adds the identity r to ids.
+func addChildren(r *Identity, ids []*Identity) []*Identity {
 	ids = appendIfNotIn(ids, r)
-	// But if it doesn't have any children, then return
-	if r.Values == nil {
-		return ids
-	}
-	// Else iterate through them
+
+	// Iterate through the values of r.
 	for _, ch := range r.Values {
-		ids = learnChildren(ch, ids)
+		ids = addChildren(ch, ids)
 	}
 	return ids
 }
 
-func findIdentityBase(baseStr string, mod *Module) (*resolvedIdentity, []error) {
-	var base *resolvedIdentity
+// findIdentityBase returns the resolved identity that is corresponds to the
+// baseStr string in the context of the Module mod.
+func (mod *Module) findIdentityBase(baseStr string) (*resolvedIdentity, []error) {
+	var base resolvedIdentity
 	var ok bool
 	var errs []error
 
@@ -83,23 +91,26 @@ func findIdentityBase(baseStr string, mod *Module) (*resolvedIdentity, []error) 
 		// This is a local identity which is defined within the current
 		// module
 		keyName := fmt.Sprintf("%s:%s", rootPrefix, baseName)
-		base, ok = identityDict.dict[keyName]
+		base, ok = identities.dict[keyName]
 		if !ok {
-			errs = append(errs, fmt.Errorf("can't resolve the local base %s as %s", baseStr, keyName))
+			errs = append(errs, fmt.Errorf("can't resolve the local base %s as %s",
+				baseStr, keyName))
 		}
+		break
 	default:
 		// This is an identity which is defined within another module
-		externalModule := FindModuleByPrefix(mod, basePrefix)
-		if externalModule == nil {
-			errs = append(errs, fmt.Errorf("can't find external module with prefix %s", basePrefix))
+		extmod := FindModuleByPrefix(mod, basePrefix)
+		if extmod == nil {
+			errs = append(errs,
+				fmt.Errorf("can't find external module with prefix %s", basePrefix))
 		}
 
 		// Run through the identities within the remote module and find the
 		// one that matches the base that we have been specified.
-		for _, rid := range externalModule.Identities() {
+		for _, rid := range extmod.Identities() {
 			if rid.Name == baseName {
-				key, _ := makeResolvedIdentity(externalModule, rid)
-				if id, ok := identityDict.dict[key]; ok {
+				key := rid.PrefixedName()
+				if id, ok := identities.dict[key]; ok {
 					base = id
 				} else {
 					errs = append(errs, fmt.Errorf("can't find base %s", baseStr))
@@ -109,29 +120,27 @@ func findIdentityBase(baseStr string, mod *Module) (*resolvedIdentity, []error) 
 		}
 		// Error if we did not find the identity that had the name specified in
 		// the module it was expected to be in.
-		if base == nil {
+		if base.isEmpty() {
 			errs = append(errs, fmt.Errorf("can't resolve remote base %s", baseStr))
 		}
 	}
-	return base, errs
+	return &base, errs
 }
 
 func (ms *Modules) resolveIdentities() []error {
-
-	// We are inplace modifying this global, and hence we have a lock on it.
-	defer identityDict.mu.Unlock()
-	identityDict.mu.Lock()
+	defer identities.mu.Unlock()
+	identities.mu.Lock()
 
 	var errs []error
 
 	// Across all modules, read the identity values that have been extracted
-	// from them, and compile them into a "fully resolved" map that mean that
+	// from them, and compile them into a "fully resolved" map that means that
 	// we can look them up based on the 'real' prefix of the module and the
 	// name of the identity.
 	for _, mod := range ms.Modules {
 		for _, i := range mod.Identities() {
-			keyName, r := makeResolvedIdentity(mod, i)
-			identityDict.dict[keyName] = r
+			keyName, r := newResolvedIdentity(mod, i)
+			identities.dict[keyName] = *r
 		}
 	}
 
@@ -139,15 +148,15 @@ func (ms *Modules) resolveIdentities() []error {
 	// fully resolved identity statement. The intention here is to make sure
 	// that the Children slice is fully populated with pointers to all identities
 	// that have a base, so that we can do inheritance of these later.
-	for _, i := range identityDict.dict {
+	for _, i := range identities.dict {
 		if i.Identity.Base != nil {
 			// This identity inherits from another identity.
 
 			root := RootNode(i.Identity)
-			base, baseErr := findIdentityBase(i.Identity.Base.asString(), root)
+			base, baseErr := root.findIdentityBase(i.Identity.Base.asString())
 
 			if baseErr != nil {
-				errs = append(errs, errs...)
+				errs = append(errs, baseErr...)
 				continue
 			}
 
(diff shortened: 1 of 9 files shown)


pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+	"fmt"
+	"sync"
+)
+
+// This file implements data structures and functions that relate to the
+// identity type.
+
+type identityDictionary struct {
+	mu   sync.Mutex
+	dict map[string]*resolvedIdentity
```
pborman: Only make this a reference if you needto.  The type is small so it is more efficient to just be a map of resolvedIdentity rather than a reference to one.

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+
+var identityDict = identityDictionary{dict: map[string]*resolvedIdentity{}}
+
+type resolvedIdentity struct {
+	Module   *Module
+	Identity *Identity
+	//Children []*resolvedIdentity
+}
+
+func makeResolvedIdentity(m *Module, i *Identity) (string, *resolvedIdentity) {
```
pborman: newResolvedIdentity

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+	r := &resolvedIdentity{
+		Module:   m,
+		Identity: i,
+	}
+	prefix := RootNode(i).GetPrefix()
+	keyName := fmt.Sprintf("%s:%s", prefix, i.Name)
+	return keyName, r
+}
+
+func appendIfNotIn(ids []*Identity, chk *Identity) []*Identity {
```
pborman: // addIdentity adds chk to ids, if not already listed, and returns the possibly new ids.
addIdentity(...) ...

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+	return append(ids, chk)
+}
+
+func learnChildren(r *Identity, ids []*Identity) []*Identity {
+	// Learn this child
+	ids = appendIfNotIn(ids, r)
+	// But if it doesn't have any children, then return
+	if r.Values == nil {
```
pborman: This check is unneeded as the range will be a no-op in this case

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+		// This is a local identity which is defined within the current
+		// module
+		keyName := fmt.Sprintf("%s:%s", rootPrefix, baseName)
+		base, ok = identityDict.dict[keyName]
+		if !ok {
+			errs = append(errs, fmt.Errorf("can't resolve the local base %s as %s", baseStr, keyName))
+		}
+	default:
+		// This is an identity which is defined within another module
+		externalModule := FindModuleByPrefix(mod, basePrefix)
```
pborman: I would probably just call this emod or extmod, typical Go style is to use pithy names.

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+		if externalModule == nil {
+			errs = append(errs, fmt.Errorf("can't find external module with prefix %s", basePrefix))
+		}
+
+		// Run through the identities within the remote module and find the
+		// one that matches the base that we have been specified.
+		for _, rid := range externalModule.Identities() {
+			if rid.Name == baseName {
+				key, _ := makeResolvedIdentity(externalModule, rid)
```
pborman: `key := fmt.Sprintf("%s:%s", RootNode(rid).GetPrefix(), rid.Name)
`
Or better yet:

```
func (*r resolvedIdenity() key() string {
    fmt.Sprintf("%s:%s", RootNode(r).GetPrefix(), r.Name)
}

```
and then above

```
func newResolvedIdentity(m *Module, i *Identity) (string, *resolvedIdentity) {
        return &resolvedIdentity {
                Module: m,
                Identity: i.key(),
        }
}

```

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+		if base == nil {
+			errs = append(errs, fmt.Errorf("can't resolve remote base %s", baseStr))
+		}
+	}
+	return base, errs
+}
+
+func (ms *Modules) resolveIdentities() []error {
+
+	// We are inplace modifying this global, and hence we have a lock on it.
```
pborman: generally you do not need this sort of comment.  The fact that you are locking it implies you are either modifying it or can't afford someone else to modify it while you are using it.

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+	defer identityDict.mu.Unlock()
+	identityDict.mu.Lock()
+
+	var errs []error
+
+	// Across all modules, read the identity values that have been extracted
+	// from them, and compile them into a "fully resolved" map that mean that
```
pborman: s/mean/means

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+// identity type.
+
+type identityDictionary struct {
+	mu   sync.Mutex
+	dict map[string]*resolvedIdentity
+}
+
+// Global dictionary of resolved identities
+
+var identityDict = identityDictionary{dict: map[string]*resolvedIdentity{}}
```
pborman: this would more normally be called "identities" (plural, not dict)

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+	// that have a base, so that we can do inheritance of these later.
+	for _, i := range identityDict.dict {
+		if i.Identity.Base != nil {
+			// This identity inherits from another identity.
+
+			root := RootNode(i.Identity)
+			base, baseErr := findIdentityBase(i.Identity.Base.asString(), root)
+
+			if baseErr != nil {
+				errs = append(errs, errs...)
```
pborman: don't you mean

`errs = append(errs, baseErr...)`

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+		base, ok = identityDict.dict[keyName]
+		if !ok {
+			errs = append(errs, fmt.Errorf("can't resolve the local base %s as %s", baseStr, keyName))
+		}
+	default:
+		// This is an identity which is defined within another module
+		externalModule := FindModuleByPrefix(mod, basePrefix)
+		if externalModule == nil {
+			errs = append(errs, fmt.Errorf("can't find external module with prefix %s", basePrefix))
+		}
```
robshakir: Yes, you're right, thanks.

pkg/yang/identity.go
robshakir: Done, thanks!

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+		return ids
+	}
+	// Else iterate through them
+	for _, ch := range r.Values {
+		ids = learnChildren(ch, ids)
+	}
+	return ids
+}
+
+func findIdentityBase(baseStr string, mod *Module) (*resolvedIdentity, []error) {
```
robshakir: Agreed, that makes more sense.

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, id := range ids {
+		if id == chk {
+			return ids
+		}
+	}
+	return append(ids, chk)
+}
+
+func learnChildren(r *Identity, ids []*Identity) []*Identity {
+	// Learn this child
```
robshakir: Thanks, pruned.

pkg/yang/identity.go
robshakir: Thanks, fixed.

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+func appendIfNotIn(ids []*Identity, chk *Identity) []*Identity {
+	for _, id := range ids {
+		if id == chk {
+			return ids
+		}
+	}
+	return append(ids, chk)
+}
+
+func learnChildren(r *Identity, ids []*Identity) []*Identity {
```
robshakir: Thanks, done.

pkg/yang/identity.go
robshakir: Thanks. Done.

pkg/yang/identity.go
robshakir: I added a `PrefixedName` function to the `Identity` type - this output is useful outside of the YANG package IMHO since it it what identities are referred to in some serialisations that I've seen. In the future we'll probably want something that has a module name prefix there too (used by IETF JSON serialisation).

pkg/yang/identity.go
robshakir: Sorry, yes, I think this comment was more useful for me than in this code!

pkg/yang/identity.go
robshakir: Fixed.

pkg/yang/identity.go
robshakir: Fixed.

pkg/yang/identity.go
robshakir: Yes, thanks!

pkg/yang/identity.go
```diff
@@ -0,0 +1,169 @@
...
+
+			// Append this value to the children of the base identity.
+			base.Identity.Values = append(base.Identity.Values, i.Identity)
+		}
+	}
+
+	// Do a final sweep through the identities to build up their children.
+	for _, i := range identityDict.dict {
+		newValues := []*Identity{}
+		for _, j := range i.Identity.Values {
```
robshakir: Yes, the issue is that for a particular identity it is only the children of the identity that are valid values for a particular identityref.

So in the construct:

```
identity FOO;
identity BAR { base FOO; }
identity BAZ { base FOO; }

leaf A {
  type identityref {
    base FOO;
  }
}
```

Only `BAR` and `BAZ` are valid values for `A`, so the intention here is to ensure that the top-level identity is not used, but then subsequent children are.

robshakir:
Thanks @pborman - I corrected all the issues that you noted thus far, PTAL. I also re-worked the test cases according to the discussion in the previous PR.

pkg/yang/identity.go
```diff
@@ -0,0 +1,178 @@
...
+
+// resolvedIdentity is an Identity that has been disambiguated.
+type resolvedIdentity struct {
+	Module   *Module
+	Identity *Identity
+}
+
+// isEmpty determines whether the resolvedIdentity struct value was defined.
+func (r resolvedIdentity) isEmpty() bool {
+	if r.Module == nil && r.Identity == nil {
```
pborman: return r.Module == nil && r.Identity == nil

pkg/yang/identity.go
```diff
@@ -0,0 +1,178 @@
...
+			errs = append(errs, fmt.Errorf("can't resolve the local base %s as %s",
+				baseStr, keyName))
+		}
+		break
+	default:
+		// This is an identity which is defined within another module
+		extmod := FindModuleByPrefix(mod, basePrefix)
+		if extmod == nil {
+			errs = append(errs,
+				fmt.Errorf("can't find external module with prefix %s", basePrefix))
```
pborman: I think you need a break after this otherwise line 110 will panic if extmod is nil.

pkg/yang/types.go
```diff
@@ -139,6 +139,16 @@ func (t *Typedef) resolve() []error {
 		y.Default = t.Default.Name
 	}
 
+	if t.Type.IdentityBase != nil {
+		// We need to copy over the IdentityBase statement if the type has one
+		if idBase, err := RootNode(t).findIdentityBase(t.Type.IdentityBase.Name); err == nil {
+			y.IdentityBase = idBase.Identity
+		} else {
+			return []error{fmt.Errorf("Could not resolve identity base for typedef: %s",
+				t.Type.IdentityBase.Name)}
```
pborman: Join this with the line above.

I used to always indent my parameters, but Go style seems to be "make a long line and use shorter names".   Even though I preferred idents in the past, now it seems easier to read without them, mainly because they are rarely used (there still are times to indent parameters, but this one is short so no reason).

pkg/yang/types_builtin.go
```diff
@@ -246,9 +246,13 @@ func (e *EnumType) ValueMap() map[int64]string {
 // refer to either a builtin type or type specified with typedef.  Not
 // all fields in YangType are used for all types.
 type YangType struct {
-	Name             string
-	Kind             TypeKind    // Ynone if not a base type
-	Base             *Type       `json:"-"` // dervied from
+	Name string
+	Kind TypeKind // Ynone if not a base type
+	// Base represents the base Type if this is a derived type
```
pborman: I still would like these to be short end-of-line comments and anything more complicated put in the description of the struct.

Some something like (not sure on wording) in the struct comment:

```
// ...  The Base field is set to the ultimate base type for derived types.  IdentityBase is the
// base statement of an Identityref type.

```
And then:

```
        Base         *Type `json:"-"` // base type dervied from
        IdentityBase *Identity        // identityref's base

```

in the struct itself.

robshakir: Sorry - I thought I'd fixed these, let me just check that I did that in the right branch. My mistake!

pkg/yang/identity_test.go
```diff
@@ -0,0 +1,415 @@
...
+		in: []inputModule{
+			inputModule{
+				name: "idtest-one",
+				content: `
+module idtest-one {
+  namespace "urn:idone";
+  prefix "idone";
+
+  identity TEST_ID;
+}
```
pborman: It is up to you, but this might be more readable if  you indent the text to be one tab in from content:

```
                        content: `
                                module idtest-one {
                                        namespace "urn:idone";
                                ...
                        `,
                },

```

pkg/yang/identity.go
robshakir: Thanks. Corrected.

pkg/yang/types.go
robshakir: Ah, interesting - I have had the 80-char-maximum drilled into me clearly :-) Removed this linebreak.

pkg/yang/types_builtin.go
robshakir: Re-fixed.

pkg/yang/identity_test.go
robshakir: ACK, done.

robshakir:
Thanks! Fixes for the above issues pushed. PTAL!

pkg/yang/types_builtin.go
pborman: You still have the `// Base represents` comment in here.

pkg/yang/identity_test.go
```diff
@@ -0,0 +1,415 @@
...
+			inputModule{
+				name: "idtest-one",
+				content: `
+					module idtest-one {
+					  namespace "urn:idone";
+					  prefix "idone";
+
+					  identity TEST_ID;
+					}
+    `},
```
robshakir: Sorry, slight editor weirds on this, PTAL.

robshakir: Looks fine - slight editor weirds on this one, PTAL and see if this looks OK now.

PR #28, opened 2016-12-17
yang: Entry: support Leaf default values

andaru:
After ToEntry is called, if a "leaf" has an explicit "default" value
in the YANG schema, that leaf's Entry will have its Default field set
to the schema value.

 * Add a Default (string) field to the Entry struct.
 * During ToEntry, for Leaf nodes, copy the "default" value to the
   new Entry Default field.
 * Provide a function DefaultValue on *yang.Entry, which returns
   either the explicit leaf default, the Entry's type's default
   if it has one or the empty string otherwise.
 * Add unit test for the DefaultValue function covering leaf usage.

### DIFF
--- a/pkg/yang/entry_test.go
+++ b/pkg/yang/entry_test.go
@@ -285,7 +285,7 @@ module defaults {
 
   typedef string-default {
     type string;
-    default "typedef default string";
+    default "typedef default value";
   }
 
   grouping common {
@@ -297,7 +297,7 @@ module defaults {
     container common-withdefault {
       leaf string {
         type string;
-        default "default string";
+        default "default value";
       }
     }
     container common-typedef-withdefault {
--- a/pkg/yang/entry_test.go
+++ b/pkg/yang/entry_test.go
@@ -17,6 +17,7 @@ package yang
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/yang/entry_test.go
```diff
@@ -264,3 +265,112 @@ func TestEntryNamespace(t *testing.T) {
...
+	}
+
+	modtext := `
+module defaults {
+  namespace "urn:defaults";
+  prefix "defaults";
+
+  typedef string-default {
+    type string;
+    default "typedef default string";
```
marcushines: *nit 
"typedef default value"

pkg/yang/entry_test.go
```diff
@@ -264,3 +265,112 @@ func TestEntryNamespace(t *testing.T) {
...
+  grouping common {
+    container common-nodefault {
+      leaf string {
+        type string;
+      }
+    }
+    container common-withdefault {
+      leaf string {
+        type string;
+        default "default string";
```
marcushines: "default value"

pkg/yang/entry_test.go
andaru: Done

pkg/yang/entry_test.go
andaru: Done

andaru:
Happy New Year! 💥  

I've addressed the comments on the literals in the tests with the newest commit.

PR #33, opened 2017-02-09
yang: Allow file names with revision-dates

andaru:
This change permits scanning YANG paths for YANG module files matching
the specification in section 5.2 of RFC6020. These include module names
with revision-date fields, such as an example module "foo", with a
revision date of 2016-12-31:

  foo@2016-12-31.yang

A module that makes an import for "foo" without specifying an
revision-date argument in its import statement (i.e., the regular state
of YANG modules found in the wild) will, before this change, fail to
find a module file (it would look for "foo.yang" in relevant paths).

After this change, goyang can successfully import the IETF RFC modules
directory (e.g,. importing ietf-system will successfully import the
latest versions of the ietf-types and ietf-netconf-acm modules).

### DIFF
--- a/pkg/yang/file.go
+++ b/pkg/yang/file.go
@@ -138,15 +138,17 @@ func findInDir(dir, name string, recurse bool) string {
 	}
 
 	for _, fi := range fis {
-		if !fi.IsDir() {
-			if fi.Name() == name {
+		switch {
+		case !fi.IsDir():
+			if fn := fi.Name(); fn == name {
 				return filepath.Join(dir, name)
-			} else if !strings.Contains(name, "@") && strings.HasPrefix(fi.Name(), mname+"@") && strings.HasSuffix(fi.Name(), ".yang") {
-				candidates = append(candidates, fi.Name())
+			} else if !strings.Contains(name, "@") {
+				// the query had no revision-date so look for candidate revisions
+				if strings.HasPrefix(fn, mname+"@") && strings.HasSuffix(fn, ".yang") {
+					candidates = append(candidates, fn)
+				}
 			}
-			continue
-		}
-		if recurse {
+		case recurse:
 			if n := findInDir(filepath.Join(dir, fi.Name()), name, recurse); n != "" {
 				return n
 			}
@@ -159,8 +161,5 @@ func findInDir(dir, name string, recurse bool) string {
 	// the case of revision-date fields per the module file name format (RFC6020
 	// section 5.2) should be the newest candidate.
 	sort.Strings(candidates)
-	if best := candidates[len(candidates)-1]; best != "" {
-		return filepath.Join(dir, best)
-	}
-	return ""
+	return filepath.Join(dir, candidates[len(candidates)-1])
 }
--- a/pkg/yang/file.go
+++ b/pkg/yang/file.go
@@ -19,6 +19,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/yang/file.go
```diff
@@ -115,22 +124,43 @@ func findFile(name string) (string, string, error) {
...
+	mname := name
+	if strings.HasSuffix(mname, ".yang") {
+		mname = name[:len(name)-len(".yang")]
+	}
+
 	for _, fi := range fis {
 		if !fi.IsDir() {
 			if fi.Name() == name {
 				return filepath.Join(dir, name)
+			} else if !strings.Contains(name, "@") && strings.HasPrefix(fi.Name(), mname+"@") && strings.HasSuffix(fi.Name(), ".yang") {
```
pborman: I think you can drop the strings.Contains as it  is implicit with the HasPrefix call.

If you are worried about the call to fi.Name(), you could make a local variable and use it throughout this block (below line 141) but I don't think that it is really necessary.

pkg/yang/file.go
```diff
@@ -115,22 +124,43 @@ func findFile(name string) (string, string, error) {
...
 		return ""
 	}
+	var candidates []string
+	mname := name
+	if strings.HasSuffix(mname, ".yang") {
+		mname = name[:len(name)-len(".yang")]
+	}
+
 	for _, fi := range fis {
 		if !fi.IsDir() {
```
pborman: This probably should be a switch statement:

switch {
case !fi.IsDir():
    ....
case recurse:
    ....
}

And then we can drop the continue

pkg/yang/file.go
```diff
@@ -115,22 +124,43 @@ func findFile(name string) (string, string, error) {
...
+	if len(candidates) == 0 {
+		return ""
+	}
+	// sort the candidates and return the ultimate ("highest") value, which in
+	// the case of revision-date fields per the module file name format (RFC6020
+	// section 5.2) should be the newest candidate.
+	sort.Strings(candidates)
+	if best := candidates[len(candidates)-1]; best != "" {
+		return filepath.Join(dir, best)
+	}
```
pborman: Why would best ever be ""?  That could only happen if fi.Name() returns "", which would not match the .yang suffice or the name@ prefix.

pkg/yang/file.go
andaru: It's not implicit, I don't believe. The first check is against `name` (i.e., if the request is for a non "revisioned" filename) wile the second check is against `fi.Name()`, the name we're iterating over, and confirms that the file (possibly) has a revision and thus is an eligible candidate.

From RFC6020 [s5.2.](https://tools.ietf.org/html/rfc6020#section-5.2):
```
   YANG modules and submodules are typically stored in files, one module
   or submodule per file.  The name of the file SHOULD be of the form:

     module-or-submodule-name ['@' revision-date] ( '.yang' / '.yin' )

   YANG compilers can find imported modules and included submodules via
   this convention.
```

So we confirm that the filename we're iterating over starts with `module-name@` and ends in `.yang`, which (perhaps with the addition of a length check to validate the `revision-date` is not empty?) is enough to satisfy the constraint described above.

pkg/yang/file.go
andaru: I agree, it shouldn't be; a left-over sanity check. I'll remove this.

pkg/yang/file.go
andaru: Nice, will do.

andaru:
Thanks for the review, Paul, I've answered and addressed your comments.

pkg/yang/file.go
andaru: Switched to using a `switch` statement and added a comment to describe the matching logic a bit better.

pborman:
Sorry for the delay.

Issue #35, opened 2017-03-15
Add handling of circular dependencies


robshakir:
In some schemas (non-OpenConfig), there exist circular dependencies between a set of submodules, based on YANG 1.0 not considering entities defined in a submodule to be accessible to another submodule of the same module without an explicit `include` statement. Currently, goyang will loop around these statements. This issue tracks adding hooks to handle such cases.

A canonical example case is the one used in (pyangbind unit tests)[https://github.com/robshakir/pyangbind/tree/nokia-fix/tests/nokia]. `mod-a` includes `subm-x` and `subm-b` which include each other.

I plan to add support for a command-line option to be able to handle this case, despite the fact that it violates the requirements of RFC6020. Comments welcome - otherwise I will open a PR.

pborman:
Having it a command line option that sets a public global variable in the yang package seems like a good avenue to take.  Thanks!

robshakir:
Thanks Paul - PR opened.

PR #36, opened 2017-03-15
Add Support for Ignoring Circular Dependencies

robshakir:
Per #35, this PR requests to add functionality to allow circular dependencies in submodules to be ignored, the default behaviour in this case is also modified to ensure that goyang doesn't enter an infinite loop in this case, and rather returns an error.

In order to control whether an error should be thrown, a `ParseOptions` struct is created which allows particular options (such as the one to ignore circular dependencies) to be set by the calling package. A command-line flag is also added to the `goyang` binary to allow this to be specified.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -10,4 +10,5 @@
 #     Name <email address>
 
 Paul Borman <borman@google.com>
+Andrew Fort <afort@arista.com>
 Rob Shakir <robjs@google.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -10,3 +10,4 @@
 #     Name <email address>
 
 Paul Borman <borman@google.com>
+Rob Shakir <robjs@google.com>
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -509,12 +509,12 @@ func ToEntry(n Node) (e *Entry) {
 				// us to ensure that we do not process circular dependencies that exist
 				// in some module's include chains.
 				switch {
-				case a.Module.NName() == n.NName() && !ParseOptions.IgnoreSubmoduleCircularDependencies:
-					e.addError(fmt.Errorf("%s: has a circular dependency", n.NName()))
-				case a.Module.NName() == n.NName():
+				case a.Module.NName() != n.NName():
+					e.merge(a.Module.Prefix, ToEntry(a.Module))
+				case ParseOptions.IgnoreSubmoduleCircularDependencies:
 					continue
 				default:
-					e.merge(a.Module.Prefix, ToEntry(a.Module))
+					e.addError(fmt.Errorf("%s: has a circular dependency", n.NName()))
 				}
 			}
 		case "leaf":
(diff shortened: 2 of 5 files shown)


CONTRIBUTORS
```diff
@@ -9,4 +9,5 @@
 # Names should be added to this file as:
 #     Name <email address>
 
-Paul Borman <borman@google.com>
\ No newline at end of file
+Paul Borman <borman@google.com>
+Rob Shakir <robjs@google.com>
```
pborman: We should add Andrew Fort <afort@arista.com> as well.

pkg/yang/entry.go
```diff
@@ -504,7 +504,18 @@ func ToEntry(n Node) (e *Entry) {
 			// available.  There is nothing else for us to do.
 		case "include":
 			for _, a := range fv.Interface().([]*Include) {
-				e.merge(a.Module.Prefix, ToEntry(a.Module))
+				// Specifically handle the case that the submodule's name is the same
+				// as the node that we are currently converting to an entry. This allows
+				// us to ensure that we do not process circular dependencies that exist
+				// in some module's include chains.
+				switch {
```
pborman: switch {
case a.Module.NName() != n.NName():
    e.merge(a.Module.Prefix, ToEntry(a.Module))
case ParseOptions.IgnoreSubmoduleCircularDependencies:
    continue
default:
    e.addError(fmt.Errorf("%s: has a circular dependency", n.NName()))
}

Makes for a bit shorter lines, a single comparison, and handles the typical case first.

pkg/yang/options.go
```diff
@@ -0,0 +1,27 @@
...
+// including specific overrides for potentially problematic YANG constructs.
+type Options struct {
+	// IgnoreSubmoduleCircularDependencies specifies whether circular dependencies
+	// between submodules. Setting this value to true will ensure that this
+	// package will explicitly ignore the case where a submodule will include
+	// itself through a circular reference.
+	IgnoreSubmoduleCircularDependencies bool
+}
+
+var ParseOptions = Options{}
```
pborman: What is the idea behind creating a structure rather than just

var IgnoreSubmoduleCircularDependencies bool

at the top level? 

If we were to plumb this through so you could be parsing different trees concurrently with differing options then a structure would make sense, but since we are not doing that, I am not sure what it offers (other than 13 more characters).

pkg/yang/entry.go
robshakir: SGTM, done.

pkg/yang/options.go
robshakir: It struck me that there were potential other errors/parsing options that we might want to ignore here, so it might be cleaner to group them together into a struct (e.g., pyangbind has a number of inputs that specifically say for extensions that need to be handled etc.). I'm happy to have individual flags though if this is preferred.

CONTRIBUTORS
robshakir: Done.

robshakir:
Thanks for the review @pborman !

pkg/yang/options.go
robshakir: (Having this struct also makes it cleaner in my calling code, just because options are handed to code generation that are passed through to the `pkg/yang` library - so having this struct lets me define a single object that is populated by a binary and the code generation library passes directly through to this library that is extensible relatively easily).

pborman: Okay, it wasn't a big deal either way :-)

PR #40, opened 2017-03-21
Add handling for circular dependencies which are not direct submodules.

robshakir:
Hi Paul,

This adds some further handling for circular dependencies that is seen in a set of modules that we're wanting to code generation for. PTAL!

Thanks again for the reviews :-)

Cheers,
r.

``` 
  * (M) pkg/yang/entry.go
    - Add handling for circular dependencies that are not simply direct
      submodule inclusions, such as those seen in some vendor YANG modules.
      This implementation uses a map of entity (submodule, or module) to
      included submodule such that grandchildren of a module include a
      submodule that was already included can be detected.
  * (M) pkg/yang/entry_test.go
    - Add testing for particular cases observed in vendor modules.
  * (M) pkg/yang/find.go
    - Ensure that FindGrouping cannot enter an infinite loop when
      circular dependencies exist - either exiting gracefully, or
      skipping already inspected submodules.
```

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -311,8 +311,9 @@ var entryCache = map[Node]*Entry{}
 
 // mergedSubmodule is used to prevent re-parsing a submodule that has already
 // been merged into a particular entity when circular dependencies are being
-// ignored.
-var mergedSubmodule = map[string]map[string]bool{}
+// ignored. The keys of the map are a string that is formed by concatenating
+// the name of the including (sub)module and the included submodule.
+var mergedSubmodule = map[string]bool{}
 
 var depth = 0
 
@@ -518,13 +519,14 @@ func ToEntry(n Node) (e *Entry) {
 				//    imports A, then we need to check whether we already have merged
 				//    the specified module during this parse attempt. We check this
 				//    against a map of merged submodules.
-				_, merged := mergedSubmodule[a.Module.NName()][n.NName()]
+				// The key of the map used is a synthesised value which is formed by
+				// concatenating the name of this node and the included submodule,
+				// separated by a ":".
+				key := n.NName() + ":" + a.Module.NName()
+				merged := mergedSubmodule[key]
 				switch {
 				case !merged && a.Module.NName() != n.NName():
-					if _, ok := mergedSubmodule[a.Module.NName()]; !ok {
-						mergedSubmodule[a.Module.NName()] = map[string]bool{}
-					}
-					mergedSubmodule[a.Module.NName()][n.NName()] = true
+					mergedSubmodule[key] = true
 					e.merge(a.Module.Prefix, ToEntry(a.Module))
 				case ParseOptions.IgnoreSubmoduleCircularDependencies:
 					continue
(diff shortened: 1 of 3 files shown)


pkg/yang/find.go
```diff
@@ -46,7 +46,7 @@ func trimPrefix(n Node, name string) string {
 // FindGrouping finds the grouping named name in one of the parent node's
 // grouping fields.  If no parent has the named grouping, nil is returned.
 // Imported and included modules are also checked.
-func FindGrouping(n Node, name string) *Grouping {
+func FindGrouping(n Node, name string, checkedInclude map[string]bool) *Grouping {
```
pborman: I think this should be called "seen" and commented above:

Seen is a list of modules previously seen by FindGrouping during traversal.

pkg/yang/entry.go
```diff
@@ -504,17 +509,27 @@ func ToEntry(n Node) (e *Entry) {
...
+				// Handle circular dependencies between submodules. This can occur in
+				// two ways:
+				//  - Where submodule A imports submodule B, and vice versa then the
+				//    whilst processing A we will also try and process A (learnt via
+				//    B). The default case of the switch handles this case.
+				//  - Where submodule A imports submodule B that imports C, which also
+				//    imports A, then we need to check whether we already have merged
+				//    the specified module during this parse attempt. We check this
+				//    against a map of merged submodules.
+				_, merged := mergedSubmodule[a.Module.NName()][n.NName()]
```
pborman: Maybe you can simplify this to a single map.  Two way I can think of:

key := a.Module.NName() + ":" + n.NName()

key := struct {
    Module string
    Include string
} { Module: a.Module.NName(), Include: n.NName()}

Obviously the second case should use a names type.

Now you can just say

merged := mergedSubmodule[key] 

If you want to use

_, merged := mergedSubmodule[key] 

then mergedSubmodule should be a map to struct{}

pkg/yang/find.go
```diff
@@ -74,15 +74,28 @@ func FindGrouping(n Node, name string) *Grouping {
...
+				if g := FindGrouping(i.Module, pname, checkedInclude); g != nil {
 					return g
 				}
 			}
 		}
 		v = e.FieldByName("Include")
 		if v.IsValid() {
 			for _, i := range v.Interface().([]*Include) {
-				if g := FindGrouping(i.Module, name); g != nil {
+				if _, c := checkedInclude[i.Module.Name]; c {
```
pborman: Either:

if seen[i.Module.Name] {

or change the map to map[string]struct{}

pkg/yang/find.go
robshakir: Done.

pkg/yang/find.go
robshakir: Done, thanks.

pkg/yang/entry.go
robshakir: Yes, that's much neater, thanks.

robshakir:
Thanks @pborman - review much appreciated.

PR #44, opened 2017-05-25
Follow io.Writer API contract on returning written size

oshothebig:
According to the [io.Writer Interface](https://golang.org/pkg/io/#Writer), returning `n` must be the number of bytes written from the given byte slice `p`, meaning 0 <= `n` <= `len(p)`.

However, `iw.Write()` returns the number of bytes written to the underlying Writer on success. The value could be larger than the length of the original byte slice because of joining.

This patch tries to fix this issue and return the number of bytes actually written from the original byte slice on error.

### DIFF
--- a/pkg/indent/indent.go
+++ b/pkg/indent/indent.go
@@ -76,7 +76,37 @@ func (w *iw) Write(buf []byte) (int, error) {
 	if !w.partial {
 		lines = append([][]byte{[]byte{}}, lines...)
 	}
-	buf = bytes.Join(lines, w.prefix)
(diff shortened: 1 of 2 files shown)


oshothebig:
Could someone look through this PR and make a response?

pborman:
Sorry, it got pushed off my queue by other internal work.  There used to be a debate on the question if Write should return the number of bytes consumed or the number of bytes written.  It appears that question is now answered, the number of bytes consumed.  Thanks for the fix.

oshothebig:
Thank you for merging this. I'm happy that I could help the project a little bit.

PR #45, opened 2017-06-13
Add support for marshaling yang.Entry to JSON.

robshakir:
```  
* (M) pkg/yang/ast.go:
  * (M) pkg/yang/parse_test.go:
  * (M) pkg/yang/parse.go:
   - Adopt changes to the Statement type such that Keyword and Argument
     are now public members of the struct. This is used in the case that
     an extension is defined to allow an external consumer to be able to
     determine these values by the value of the fields rather than the
     helper methods. Allows this information to be marshalled to JSON.
  * (M) pkg/yang/entry.go:
   - Add JSON annotations to the entry struct to ensure that there are
     no loops when marshalling an Entry. Explicitly excludes the Parent
     field since this results in circular references
     (parent.Dir[child].Parent == parent). Adds omitempty annotations to
     reduce JSON size.
  * (M) pkg/yang/types_builtin.go:
   - Add annotations to the YangType struct such that it can be
     marshalled to JSON.
  * (M) pkg/yang/yang.go:
   - Add annotations to the Value struct such that it can be marshalled.
  * (M) pkg/yang/marshal_test.go
   - Add test cases for marshalling to JSON.
```

### DIFF
--- a/pkg/yang/parse.go
+++ b/pkg/yang/parse.go
@@ -48,9 +48,9 @@ type parser struct {
 // A Statement is a generic YANG statement.  A Statement may have optional
 // sub-statement (i.e., a Statement is a tree).
 type Statement struct {
-	keyword     string
+	Keyword     string
 	hasArgument bool
-	argument    string
+	Argument    string
 	statements  []*Statement
 
 	file string
@@ -61,7 +61,7 @@ type Statement struct {
 // FakeStatement returns a statement filled in with keyword, file, line and col.
 func FakeStatement(keyword, file string, line, col int) *Statement {
 	return &Statement{
-		keyword: keyword,
+		Keyword: keyword,
 		file:    file,
 		line:    line,
 		col:     col,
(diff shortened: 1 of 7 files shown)


pkg/yang/parse_test.go
```diff
@@ -105,7 +105,7 @@ foo "\\ \S \n";
 pattern "\\ \S \n";
 `,
 			out: []*Statement{
-				SA("pattern", `\\ \S 
+        SA("pattern", `\\ \S
```
pborman: Did gofmt really put this here?  Seems like it should still be shifted over.

pkg/yang/parse.go
```diff
@@ -48,9 +48,9 @@ type parser struct {
 // A Statement is a generic YANG statement.  A Statement may have optional
 // sub-statement (i.e., a Statement is a tree).
 type Statement struct {
-	keyword     string
+	Keyword     string
 	hasArgument bool
```
pborman: Doesn't hasArgument also need to be exported?

pkg/yang/parse_test.go
robshakir: Ah, apologies, I had issues with an editor removing other whitespace, and forgot to subsequently run `gofmt`. Fixed.

pkg/yang/parse.go
robshakir: Yes, I think it does. Making this change.

PR #47, opened 2017-06-22
yang: Add support for YANG 1.1 "anydata" statement

### LINKED ISSUE #46: Support for YANG 1.1 "anydata" statements
The `anydata` statement is defined in YANG 1.1 [here](https://tools.ietf.org/html/rfc7950#section-7.10).

While `goyang`s validation (performed by `ToEntry`) is largely YANG 1.0 specific, the input lexer supports reading YANG 1.1 modules as they don't change the syntax dramatically (notable exception: `uses` allows more than 1 `augment` child in YANG 1.1) and this type is used by IETF standard modules, which are increasingly using YANG 1.1.

Goyang should support lexing and handling `anydata` statements as it does `anyxml` presently.

andaru:
Support reading the YANG 1.1 "anydata" statement.

The "anydata" statement is defined at:
https://tools.ietf.org/html/rfc7950#section-7.10

Prior to this change, the presence of an "anydata"
sub-statement in YANG modules leads to errors such as:

```
ietf-restconf.yang:196:9: unknown list field: anydata
ietf-yang-patch.yang:263:9: unknown list field: anydata
```

The "anydata" statement is a YANG data node statement, similar
in nature to the YANG 1.0 "anyxml" statement. It stores an
opaque value whose use and encoding is agreed upon apriori
between parties using the field.

Add lexing tests for the "anydata" and "anyxml" statements.

Resolves #46.

### DIFF
--- a/pkg/yang/yang.go
+++ b/pkg/yang/yang.go
@@ -105,6 +105,7 @@ type Module struct {
 	Parent     Node         `yang:"Parent,nomerge"`
 	Extensions []*Statement `yang:"Ext"`
 
+	Anydata      []*AnyData      `yang:"anydata"`
 	Anyxml       []*AnyXML       `yang:"anyxml"`
 	Augment      []*Augment      `yang:"augment"`
 	BelongsTo    *BelongsTo      `yang:"belongs-to,required=submodule,nomerge"`
@@ -335,6 +336,7 @@ type Container struct {
 	Parent     Node         `yang:"Parent,nomerge"`
 	Extensions []*Statement `yang:"Ext"`
 
+	Anydata     []*AnyData   `yang:"anydata"`
 	Anyxml      []*AnyXML    `yang:"anyxml"`
 	Choice      []*Choice    `yang:"choice"`
 	Config      *Value       `yang:"config"`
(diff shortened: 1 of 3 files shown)


pkg/yang/yang.go
```diff
@@ -552,13 +557,40 @@ func (s *AnyXML) NName() string         { return s.Name }
 func (s *AnyXML) Statement() *Statement { return s.Source }
 func (s *AnyXML) Exts() []*Statement    { return s.Extensions }
 
+// An AnyData is defined in: http://tools.ietf.org/html/rfc7950#section-7.10
+//
+// AnyData are only expected in YANG 1.1 modules (those with a
+// "yang-version 1.1;" statement in the module).
+type AnyData struct {
```
andaru: Alternative: as `anydata` have the same sub-statements as `anyxml` in NETCONF 1.0/1.1, this could alternatively be written simply as;

```go
type AnyData AnyXML

func (AnyData) Kind() string { return "anydata" }
func (s *AnyData) ParentNode() Node { return s.Parent }
...
```

Do you have a preference for one or the other?

pborman: I think enumerating it is fine, they may diverge in the future.

If you did want to do the type thing, I would make AnyData be the defined one and AnyXML be the derived one.  Too bad we cant just say:

type AnyXML struct {
    AnyData
}

And then just define Kind

But that would mess up parsing.

PR #49, opened 2017-06-22
yang: Add support for YANG 1.1 "action" statement

### LINKED ISSUE #48: Support for YANG 1.1 "action" statements
In YANG 1.1 (RFC7950), `action` statements are like `rpc` elements which apply to a specific container(s) or list(s) (plurality via grouping application).

`goyang` should support lexing `action` statements and importing them to `*Entry`, similarly to `rpc` statements. However, while `rpc` statements exist only as direct children of the `module` statement, `action` may appear in the schema as direct children of the `container`, `list`, `grouping`, or `augment` statements.

Currently, `goyang` fails to load modules with `action` statements, issuing errors such as:

```
ietf-routing@2016-11-04.yang:447:9: unknown list field: action
ietf-keychain.yang:129:9: unknown list field: action
```

andaru:
Support YANG 1.1's "action" statement in goyang.

The "action" statement is defined at:
https://tools.ietf.org/html/rfc7950#section-7.10

Prior to this change, the presence of an "action"
sub-statement in YANG modules leads to errors such as:

```
ietf-routing@2016-11-04.yang:447:9: unknown list field: action
ietf-keychain.yang:129:9: unknown list field: action
```

The "action" statement has the same sub-statements as the "rpc"
statement. The difference is that instead of defining an RPC
action found only at the module level, "action" can appear as
sub-statements to "container" and "list" data nodes, as well as
"grouping" and "augment" statements similarly.

This change does the following:
  - Define the Action type (in yang.go), like an RPC.
  - Add Action as a 0..n cardinality child field to:
    Container, List, Grouping and Augment types.
  - Add tests for successful and unsuccessful module lexing and
    ToEntry parsing of RPC and Action types.

Resolves #48.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -484,6 +484,10 @@ func ToEntry(n Node) (e *Entry) {
 			if v := fv.Interface().(*Value); v != nil {
 				e.Prefix = v
 			}
+		case "action":
+			for _, r := range fv.Interface().([]*Action) {
+				e.add(r.Name, ToEntry(r))
+			}
 		case "augment":
 			for _, a := range fv.Interface().([]*Augment) {
 				ne := ToEntry(a)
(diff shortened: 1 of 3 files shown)


robshakir:
LGTM!

PR #52, opened 2017-06-28
Handle mismatched import prefixes in Entry.Find()

robshakir:
```
 * (M) pkg/yang/entry.go
   - In YANG we can have module A, defining prefix A, imported by
     module B which imports it using its own prefix FOOBAR. In this
     case then Find() for an Entry would assume that it can use
     FOOBAR to find A within the module set that was being worked on.
     This change catalogues the imports of module B, such that there
     is a mapping between FOOBAR->A, and looks up the module using
     the prefix A.
 * (M) pkg/yang/entry_test.go
  - Enhance unit test coverage for Find, including the above case.
    We are a little more liberal than the spec says that we can be
    such that we do not error if there is /foo:bar/baz and baz
    exists within the module prefixed foo.
```

### DIFF
--- a/pkg/yang/entry_test.go
+++ b/pkg/yang/entry_test.go
@@ -1082,6 +1082,7 @@ func TestEntryFind(t *testing.T) {
 
 					leaf ctx { type string; }
 					leaf other { type string; }
+					leaf conflict { type string; }
 				}`,
 			"foo.yang": `
 				module foo {
@@ -1091,6 +1092,8 @@ func TestEntryFind(t *testing.T) {
 					container bar {
 						leaf baz { type string; }
 					}
+
+					leaf conflict { type string; }
 				}`,
 			"bar.yang": `
 				module bar {
@@ -1100,6 +1103,8 @@ func TestEntryFind(t *testing.T) {
 					container fish {
 						leaf chips { type string; }
 					}
+
+					leaf conflict { type string; }
 				}`,
 		},
 		inBaseEntryPath: "/test/ctx",
@@ -1113,6 +1118,11 @@ func TestEntryFind(t *testing.T) {
 			"/foo:bar/baz": "/foo/bar/baz",
 			// With mismatched prefixes.
 			"/baz:fish/baz:chips": "/bar/fish/chips",
+			// With conflicting node names
+			"/conflict":     "/test/conflict",
+			"/foo:conflict": "/foo/conflict",
+			"/baz:conflict": "/bar/conflict",
+			"/t:conflict":   "/test/conflict",
 		},
 	}}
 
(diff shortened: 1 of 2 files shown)


pkg/yang/entry_test.go
```diff
@@ -1012,3 +1012,143 @@ func TestActionRPC(t *testing.T) {
...
+				}`,
+			"foo.yang": `
+				module foo {
+					prefix "foo"; // matches the import above
+					namespace "urn:foo";
+
+					container bar {
+						leaf baz { type string; }
+					}
+				}`,
```
pborman: What would happen if all modules contained a container or leaf of the same name.  For example, they all have "leaf conflict { type string }".  We should still be able to get to each of them by saying conflict, foo:confclit, bar:conflict, baz:conflict.

pkg/yang/entry_test.go
```diff
@@ -1012,3 +1012,143 @@ func TestActionRPC(t *testing.T) {
...
+						leaf chips { type string; }
+					}
+				}`,
+		},
+		inBaseEntryPath: "/test/ctx",
+		wantEntryPath: map[string]string{
+			// Check we can still do intra module lookups
+			"../other":         "/test/other",
+			"/other":           "/test/other",
+			"/foo:bar/foo:baz": "/foo/bar/baz",
```
pborman: I probably have forgotten, but why can you case /other and not /test/other?  Or is other really at the root and the result is simply prepending the module name above the modules root?

andaru: I think that's because the right hand side is the `(*yang.Entry).Path()` result, which prefixes the module name.

pborman: Thanks.  I should have know that :-)

robshakir: Yep, that's right :-)

pkg/yang/entry_test.go
robshakir: I added a test case for this - it works as one would expect. Essentially, we resolve `conflict` within our own module, we resolve `our-own-prefix:conflict` within our own module, and the remote `conflict` nodes according to the local module's `prefix` for import.

pborman: Did you push it?  I don't seem to see the new code yet.

robshakir:
https://github.com/openconfig/goyang/pull/52/commits/c41f56593196bfaa32b5f73b3bebe3bbd17205c9
seems
to show it - I added conflict leaves to each module, and the corresponding
test cases in the existing multi-module test.

LMK if it doesn't show up here - it should be at
https://github.com/robshakir/goyang/blob/augment/pkg/yang/entry_test.go#L1122-L1125
 too.

Thanks,
r.

pborman:
Thanks, I see it now.

PR #53, opened 2017-07-12
Add decimal support and more unit tests.

ostromart:
Add support for YANG decimal, described in [Section 9.3](https://tools.ietf.org/html/rfc6020#section-9.3) of RFC 6020.
Number is updated with additional pointer fields to support the decimal type. The presence of these fields indicates that the number is a decimal type. 
The internal representation of decimal is int64 since this has enough bits to support the required precision for decimal.

### DIFF
--- a/pkg/yang/types_builtin.go
+++ b/pkg/yang/types_builtin.go
@@ -20,6 +20,7 @@ package yang
 import (
 	"errors"
 	"fmt"
+	"math"
 	"sort"
 	"strconv"
 	"strings"
@@ -450,12 +451,14 @@ var (
 )
 
 const (
-	MaxInt64        = 1<<63 - 1 // maximum value of a signed int64
-	MinInt64        = -1 << 63  // minimum value of a signed int64
-	AbsMinInt64     = 1 << 63   // the absolute value of MinInt64
-	MaxEnum         = 1<<31 - 1 // maximum value of an enum
-	MinEnum         = -1 << 31  // minimum value of an enum
-	MaxBitfieldSize = 1 << 32   // maximum number of bits in a bitfield
+	MaxInt64                = 1<<63 - 1        // maximum value of a signed int64
+	MinInt64                = -1 << 63         // minimum value of a signed int64
+	AbsMinInt64             = 1 << 63          // the absolute value of MinInt64
+	MaxEnum                 = 1<<31 - 1        // maximum value of an enum
+	MinEnum                 = -1 << 31         // minimum value of an enum
+	MaxBitfieldSize         = 1 << 32          // maximum number of bits in a bitfield
+	MaxFractionDigits uint8 = 18               // maximum fractional digits per Section 9.3.4
+	MaxFloat64              = float64(1 << 48) // maximum absolute float64 that can be converted to decimal
 )
 
 type NumberKind int
@@ -467,28 +470,64 @@ const (
 	MaxNumber                    // Number is maximum value allowed for range
 )
 
-// A Number is in the range of [-(1<<64) - 1, (1<<64)-1].  This range
-// is the union of uint64 and int64.
-// to indicate
+// A Number is either an integer the range of [-(1<<64) - 1, (1<<64)-1], or a
+// YANG decimal conforming to https://tools.ietf.org/html/rfc6020#section-9.3.4.
 type Number struct {
-	Kind  NumberKind
+	// Kind is the kind of number (+/-ve, min/max). Applies to both integer and
+	// decimal numbers.
+	Kind NumberKind
+	// Value is the absolute value of the number, if it is an integer.
 	Value uint64
+	// DecimalValue is the signed value of the number if it is a decimal, or nil
+	// if it is not.
+	// Decimals are represented as int64 where FractionDigits indicates the
+	// number of fractional digits.
+	DecimalValue *int64 `json:",omitempty"`
+	// Number of fractional digits in DecimalValue, if DecimalValue is non-nil.
+	FractionDigits *uint8 `json:",omitempty"`
 }
 
 var maxNumber = Number{Kind: MaxNumber}
 var minNumber = Number{Kind: MinNumber}
+
+// IsDecimal reports whether n is a decimal number.
+func (n Number) IsDecimal() bool {
+	return n.DecimalValue != nil
+}
 
 // FromInt creates a Number from an int64.
 func FromInt(i int64) Number {
 	if i < 0 {
 		return Number{Kind: Negative, Value: uint64(-i)}
 	}
-	return Number{Value: uint64(i)}
+	return Number{Kind: Positive, Value: uint64(i)}
 }
 
 // FromUint creates a Number from a uint64.
 func FromUint(i uint64) Number {
-	return Number{Value: i}
+	return Number{Kind: Positive, Value: i}
+}
+
+// FromFloat creates a Number from a float64. Input values with absolute value
+// larger than MaxFloat64 are converted into maxNumber or minNumber.
+func FromFloat(f float64) Number {
+	if f > MaxFloat64 {
+		return maxNumber
+	}
+	if f < -MaxFloat64 {
+		return minNumber
+	}
+	var fracDig uint8
+	for ; Frac(f) != 0.0 && fracDig <= MaxFractionDigits; fracDig++ {
+		f *= 10.0
+	}
+	v := int64(f)
+
+	n := Number{Kind: Positive, DecimalValue: Int64(v), FractionDigits: Uint8(fracDig)}
+	if v < 0 {
+		n.Kind = Negative
+	}
+	return n
 }
 
 // ParseNumber returns s as a Number.  Numbers may be represented in decimal,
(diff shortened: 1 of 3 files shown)


pkg/yang/types_builtin.go
```diff
@@ -506,32 +545,98 @@ func ParseNumber(s string) (n Number, err error) {
...
 	return n, err
 }
 
+// DecimalValueFromString returns an int64 representation of numStr and uint8
+// representation of the number of its fractional digits.
+// If fracDigRequired is >= 0, the number is represented with fracDigRequired
+// fractional digits, regardless of the precision of numStr, otherwise the
+// precision of numStr is used to set the number of fractional digits.
+// numStr must conform to Section 9.3.
+func DecimalValueFromString(numStr string, fracDigRequired int) (n *int64, fracDig *uint8, err error) {
```
pborman: Just have this return a Number

pkg/yang/types_builtin.go
```diff
@@ -506,32 +545,98 @@ func ParseNumber(s string) (n Number, err error) {
...
+	s := numStr
+	decCount := strings.Count(s, ".")
```
pborman: x := strings.Index(s, ".")
if x < 0 {not a number}

fd = len(s) - 1 - x
s = s[:x] + s[x+1:]
u, err := strconv.ParseUint(s)

if fd < fracDigRequired {
    m := 1
    for fd < fracDigRequired {
        m *= 10
        fd++
    }
    u *= m (probably should check for overflow)
}
if fd < fracDigRequired {
    d := 1
    for fd > fracDigRequired {
        d /= 10
        fd--
    }
    u /= d (probably should do rounding)
}

pkg/yang/types_builtin.go
```diff
@@ -506,32 +545,98 @@ func ParseNumber(s string) (n Number, err error) {
...
+			return t[:len(t)-fd] + "." + t[len(t)-fd:]
+		}
+		return t
+	}
+	return out + strconv.FormatUint(n.Value, 10)
 }
 
-// Int returns n as an int64.  It returns an error if n overflows an int64.
+// DebugString returns n's internal represenatation as a string.
+func (n Number) DebugString() string {
```
pborman: Is this used?  Now that we don't have pointers, fmt.Sprint(n) is probably sufficient (e.g., %v)

pkg/yang/types_builtin.go
```diff
@@ -506,32 +545,98 @@ func ParseNumber(s string) (n Number, err error) {
...
+		return fmt.Sprintf("{%v, %d, %d, %d}", n.Kind, n.Value, *n.DecimalValue, *n.FractionDigits)
+	}
+	return fmt.Sprintf("{%v, %d, %p, %p}", n.Kind, n.Value, n.DecimalValue, n.FractionDigits)
+}
+
+// Int returns n as an int64.  It returns an error if n overflows an int64 or
+// the number is decimal.
 func (n Number) Int() (int64, error) {
+	if n.IsDecimal() {
+		return 0, errors.New("cannot call Int() on floating point number")
```
pborman: on decimal64 number

It is fixed point, not floating point :-)

pkg/yang/types_builtin.go
```diff
@@ -588,17 +701,30 @@ func (n Number) Less(m Number) bool {
 		return true
 	case n.Kind != Negative && m.Kind == Negative:
 		return false
-	case n.Kind == Negative:
+	}
+
+	if n.IsDecimal() != m.IsDecimal() {
+		panic(fmt.Sprintf("Less with decimal vs. integer (%v, %v) not supported", m, n))
+	}
```
pborman: I think we should be able to compare integers with decimal64 values.

pkg/yang/types_builtin.go
```diff
@@ -588,17 +701,30 @@ func (n Number) Less(m Number) bool {
 		return true
 	case n.Kind != Negative && m.Kind == Negative:
 		return false
-	case n.Kind == Negative:
+	}
```
pborman: I think we can add:

case n.Kind == Negative && m.Kind != Negative:
    return true

pkg/yang/types_builtin.go
```diff
@@ -450,12 +451,14 @@ var (
...
-	MinEnum         = -1 << 31  // minimum value of an enum
-	MaxBitfieldSize = 1 << 32   // maximum number of bits in a bitfield
+	MaxInt64                = 1<<63 - 1        // maximum value of a signed int64
+	MinInt64                = -1 << 63         // minimum value of a signed int64
+	AbsMinInt64             = 1 << 63          // the absolute value of MinInt64
+	MaxEnum                 = 1<<31 - 1        // maximum value of an enum
+	MinEnum                 = -1 << 31         // minimum value of an enum
+	MaxBitfieldSize         = 1 << 32          // maximum number of bits in a bitfield
+	MaxFractionDigits uint8 = 18               // maximum fractional digits per Section 9.3.4
+	MaxFloat64              = float64(1 << 48) // maximum absolute float64 that can be converted to decimal
```
ostromart: Just pulled that out of the air :-) I didn't want to assume 53 bits but something more conservative. But this is a moot point since I'm getting rid of it altogether.

pkg/yang/types_builtin.go
```diff
@@ -467,28 +470,64 @@ const (
...
+	Kind NumberKind
+	// Value is the absolute value of the number, if it is an integer.
 	Value uint64
+	// DecimalValue is the signed value of the number if it is a decimal, or nil
+	// if it is not.
+	// Decimals are represented as int64 where FractionDigits indicates the
+	// number of fractional digits.
+	DecimalValue *int64 `json:",omitempty"`
+	// Number of fractional digits in DecimalValue, if DecimalValue is non-nil.
+	FractionDigits *uint8 `json:",omitempty"`
```
ostromart: Agreed, good idea.

pkg/yang/types_builtin.go
ostromart: I think that check was there already. I just removed the unnecessary case which is covered by the preceding cases.

pkg/yang/types_builtin.go
ostromart: Done.

pkg/yang/types_builtin.go
```diff
@@ -588,17 +696,36 @@ func (n Number) Less(m Number) bool {
...
+	return !n.Less(m) && !m.Less(n)
+}
+
+// toDecimal returns n represented internally with fracDigits fractional digits.
+func (n Number) toDecimal(fracDigits uint8) Number {
+	if fracDigits > MaxFractionDigits {
+		return Number{Kind: n.Kind, Value: 0, FractionDigits: MaxFractionDigits}
+	}
+
+	return Number{Kind: n.Kind, Value: n.Value * uint64(math.Pow10(int(fracDigits-n.FractionDigits))), FractionDigits: fracDigits}
```
pborman: This will always set the value to 0 if n.FractionDigits is greater than fracDigits.  You need to use the absolute value and then divide in that case.  See https://play.golang.org/p/Bz0jsgj0ae

Please first add tests that fail with the existing code and then fix the code so the tests pass.

pkg/yang/types_builtin.go
```diff
@@ -588,17 +696,36 @@ func (n Number) Less(m Number) bool {
...
 		return false
-	case n.Kind == Negative:
-		return n.Value > m.Value
+	}
+
+	// Normalize into a common representation dictated by the largest number of
+	// fraction digits in either number. Values can be compared directly in this
+	// representation.
+	maxfd := max(n.FractionDigits, m.FractionDigits)
+	nd, md := n.toDecimal(maxfd), m.toDecimal(maxfd)
```
pborman: This will fail for large number being upscaled.  See https://play.golang.org/p/Bz0jsgj0ae

Please first add tests that fail  and then fix the code so the tests pass.

pkg/yang/types_builtin.go
pborman: Oh, probably %#v since String is already defined, %v is going to call String while %#v will not.

pkg/yang/types_builtin.go
```diff
@@ -506,32 +541,97 @@ func ParseNumber(s string) (n Number, err error) {
...
+// DecimalValueFromString returns a decimal Number representation of inStr.
+// If fracDigRequired is >= 0, the number is represented with fracDigRequired
+// fractional digits, regardless of the precision of numStr, otherwise the
+// precision of numStr is used to set the number of fractional digits.
+// numStr must conform to Section 9.3.4.
+func DecimalValueFromString(numStr string, fracDigRequired int) (n Number, err error) {
+	s := numStr
+	dx := strings.Index(s, ".")
+	fracDig := 0
+	if dx > 0 {
```
pborman: Don't you want dx >= 0 so you can accept ".123"?

pkg/yang/types_builtin.go
```diff
@@ -506,32 +541,97 @@ func ParseNumber(s string) (n Number, err error) {
...
+		s = s[:dx] + s[dx+1:]
+	}
+	if fracDigRequired < 0 {
+		fracDigRequired = fracDig
+	}
+	if fracDig > fracDigRequired {
+		return n, fmt.Errorf("%s has too much precision, expect <= %d fractional digits", s, fracDigRequired)
+	}
+
+	for i := 0; i < fracDigRequired-fracDig; i++ {
```
pborman: s += "000000000000000000"[:fracDigRequired-fracDig]

There has to be at least 18 0's there.

ostromart: How do you think of this stuff? Does it come to you in a dream? :-)

pkg/yang/types_builtin.go
ostromart: I don't think there was a way to save that approach due to the overflow issue. I tried something a little different instead so toDecimal is gone.

pkg/yang/types_builtin.go
ostromart: %#v gives a weird output, I went back to enumerating the fields.

pkg/yang/types_builtin.go
ostromart: function is gone now.

pkg/yang/types_builtin.go
```diff
@@ -794,3 +930,17 @@ func (r YangRange) Contains(s YangRange) bool {
...
+}
+
+// pow10 returns 10^e without checking for overflow.
+func pow10(e uint8) uint64 {
+	var out uint64 = 1
+	for i := uint8(0); i < e; i++ {
+		out *= 10
+	} 
+	return out
+}
```
pborman: var pow10 = []uint64{
1,
10,
100,
1000,
10000,
100000,
1000000,
10000000,
100000000,
1000000000,
10000000000,
100000000000,
1000000000000,
10000000000000,
100000000000000,
1000000000000000,
10000000000000000,
100000000000000000,
1000000000000000000,
}

And you can saw pow10[n.FractionDigits]

(You don't have to do this, but I though O(1) was kinda fun :-)

pkg/yang/types_builtin.go
```diff
@@ -588,17 +699,42 @@ func (n Number) Less(m Number) bool {
...
+func (n Number) Trunc() uint64 {
+	nv := n.Value
+	e := uint64(pow10(n.FractionDigits))
+	return nv / e
+}
+
+// Frac returns the fractional part of abs(n) as a signed integer.
+func (n Number) Frac() uint64 {
+	nv := n.Value
+	e := uint64(pow10(n.FractionDigits))
```
pborman: Drop the now unnecessary typecast (and in the other places)

pkg/yang/types_builtin.go
```diff
@@ -588,17 +699,42 @@ func (n Number) Less(m Number) bool {
...
 	case n.Kind != Negative && m.Kind == Negative:
 		return false
-	case n.Kind == Negative:
-		return n.Value > m.Value
-	default:
-		return n.Value < m.Value
 	}
+
+	nt, mt := n.Trunc(), m.Trunc()
+	nf, mf := n.Frac(), m.Frac()
```
pborman: Might as well move this inside the the nt == mt block since we only need it then.

pkg/yang/types_builtin.go
pborman: Move this into the _test file since it is only needed for tests.  There is no reason I can think of that we want to generally export this.

pkg/yang/types_builtin.go
ostromart: Done.

pkg/yang/types_builtin.go
ostromart: There's only one other place I can see, am I missing something?

pkg/yang/types_builtin.go
ostromart: Neat idea but for the sake of compactness I think I'll keep it as is, since performance is not critical here.

PR #54, opened 2017-07-13
Add Entry type checking utilities.

ostromart:
These functions have been in use throughout the ocstructs and validation libraries but really belong with goyang Entry type.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -110,6 +110,13 @@ type RPCEntry struct {
 	Output *Entry
 }
 
+// A ListAttr is associated with an Entry that represents a List node
+type ListAttr struct {
+	MinElements *Value // leaf-list or list MUST have at least min-elements
+	MaxElements *Value // leaf-list or list has at most max-elements
+	OrderedBy   *Value // order of entries determined by "system" or "user"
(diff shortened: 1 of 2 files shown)


pkg/yang/entry.go
```diff
@@ -110,6 +110,13 @@ type RPCEntry struct {
 	Output *Entry
 }
 
+// A ListAttr is associated with an Entry that represents a List node
+type ListAttr struct {
+	MinElements *Value // leaf-list or list MUST have at least min-elements
+	MaxElements *Value // leaf-list or list has at most max-elements
+	OrderedBy   *Value // order of entries determined by "system" or "user"
+}
+
```
pborman: I assume there will be another pull request that actually uses this type?

ostromart: You mean ListAttr or the below functions? ListAttr is just moved to get it out of the way to Modules(). The functions below are used in ocstructs which is slated for OS release soon - that's where the change will be made to use the functions here instead.

pborman: Ahh, yes, I missed the "deleted lines" in the left column (am too used to seeing moved lines marked as moved).

PR #55, opened 2017-07-14
Fix fraction compare bug in goyang decimal.

ostromart:
Fix fraction compare bug in goyang decimal.

This addresses an issue when two decimals with fractional part are compared and they have different number of fractional leading zeros e.g. 1.05 vs 1.1. Since Frac() returned the fractional part without leading zeros, the fractional comparison using this function was wrong in these cases.


pkg/yang/types_builtin.go
```diff
@@ -724,11 +724,23 @@ func (n Number) Trunc() uint64 {
...
-// Frac returns the fractional part of abs(n) as a signed integer.
-func (n Number) Frac() uint64 {
+// normalizedFrac returns the fractional part of abs(n) as a signed integer,
+// normalized to the same number of decimal places as other e.g.
+//   n, m := 1.1, 1.05
+//   n.normalizedFrac(m) == 10
+//   m.normalizedFrac(n) == 5
+func (n Number) normalizedFrac(other Number) uint64 {
+	m := uint64(1)
+	d := int(other.FractionDigits) - int(n.FractionDigits)
```
pborman: This function works, but I think it is more complicated than need be, in particular, since you are calling it in both directions above.  I think you could  just have a simple function:

// frac returns the fraction part with a precision of 18 fractional digits.
// E.g. if n is 3.1 then n.frac() returns 100,000,000,000,000,000
func (n Number) frac() uint64 {
        frac := n.FractionDigits
        i := n.Trunc() * pow10(frac)
        return (n.Value - i) * pow10(uint8(18-frac))
}

And above say:

if nt == mt {
    return n.frac() < m.frac()
}

ostromart: Good idea. Thanks.

PR #56, opened 2017-07-17
Fix up Number parsing, displaying and fix build error.

pborman:
marshal_test.go was including pretty from the internal Google repository.

Parsing the number "0.123" resulted in 0.083 as the leading 0 cause
the parse to be in octal.

Displaying the string ".1" into a Decimal number paniced.

Display "0.1" not ".1"

### DIFF
--- a/pkg/yang/number_test.go
+++ b/pkg/yang/number_test.go
@@ -56,6 +56,10 @@ func TestNumberParse(t *testing.T) {
 		numString: ".123",
 		want:      Number{Kind: Positive, Value: 123, FractionDigits: 3},
 	}, {
+		desc:      "+ve float, leading 0",
+		numString: "0.123",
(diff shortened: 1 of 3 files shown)


pkg/yang/number_test.go
```diff
@@ -56,6 +56,10 @@ func TestNumberParse(t *testing.T) {
 		numString: ".123",
 		want:      Number{Kind: Positive, Value: 123, FractionDigits: 3},
 	}, {
+		desc:      "+ve float, leading 0",
+		numString: "0.123",
+		want:      Number{Kind: Positive, Value: 123, FractionDigits: 3},
+	}, {
```
ostromart: Would be good to add .0123 case.

pborman: Done

PR #60, opened 2017-07-21
Find included identities. 

pborman:
Identities included in submodules were not be resolved.  Make sure we hoist up submodule identities.

Fix the goyang command to exit after printing usage.

Chanage --path to do a recursive include of all directories that contain yang modules below the specified directory.

### DIFF
--- a/yang.go
+++ b/yang.go
@@ -91,7 +91,8 @@ func main() {
 
 	var traceP string
 	var help bool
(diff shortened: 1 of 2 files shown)


yang.go
```diff
@@ -146,6 +148,15 @@ Formats:
 		stop(0)
 	}
 
+	for _, path := range paths {
+		paths, err := yang.PathsWithModules(path)
```
andaru: FWIW: the "shadow" linter will be happier if you used a different name than `paths`, as you're shadowing the `paths` variable declared on line 94.

PR #63, opened 2017-08-11
Resolve bug with namespace identification for augmented nodes

robshakir:
According to RFC6020 when we have an `augment` statement then the namespace of nodes that are augmented is that of the augment*ing* module rather than the augment*ed* module. This PR fixes an issue whereby this was not correctly reported by `Namespace()` on an `Entry`.

Additionally, it adds a `InstantiatingModule` method to the `Entry` type such that one can determine the module that instantiated the leaf, for use in RFC7951 JSON serialisation.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -913,6 +913,24 @@ func (e *Entry) Namespace() *Value {
 	return new(Value)
 }
 
+// InstantiatingModule returns the YANG module which instanitated the Entry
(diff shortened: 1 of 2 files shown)


pborman:
Either you or Andrew can merge.

robshakir:
Thanks Paul!

andaru:
Thanks, Paul, I'd run into this again this week - thanks for freeing up my weekend :)

PR #64, opened 2017-08-25
Entry: DefaultValue improvement

andaru:
An Entry's `DefaultValue()` is constrained by RFC6020 section 7.6.1.,
about a leaf's default value. The final paragraph of section 7.6.1.
says:

```
   If a leaf has a "default" statement, the leaf's default value is
   the value of the "default" statement.  Otherwise, if the leaf's
   type has a default value, and the leaf is not mandatory, then the
   leaf's default value is the type's default value.  In all other
   cases, the leaf does not have a default value.
```

This change implements the second sentence in the above paragraph,
namely only using the typedef's default value when the entry is a
leaf and is not mandatory.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -1079,7 +1079,11 @@ func (e *Entry) DefaultValue() string {
 	if len(e.Default) > 0 {
 		return e.Default
 	} else if typ := e.Type; typ != nil {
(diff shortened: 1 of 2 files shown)


andaru:
Closing unmerged to correct without an additional commit.

PR #68, opened 2017-09-20
Add coverage checks via coveralls.io

robshakir:
This PR adds support for test coverage via coveralls.io.


robshakir:
I think this should work in Travis-CI, I'll update if it doesn't. It just seemed nice to have this coverage across our Go projects under OpenConfig.

Thanks for reviewing!

PR #69, opened 2017-10-18
yang: Fix Find() to retrieve rpc/action Entry

andaru:
Goyang does not successfully `(*yang.Entry).Find()` valid absolute-schema-node-identifier used in
IETF standard YANG modules, e.g.,

  `/nc:get-config/nc:input`

Such as in the `ietf-netconf-time@2016-01-26.yang` module,
where "nc" refers to ietf-netconf. This is a valid YANG
`absolute-schema-node-identifier` referring to the NETCONF
`get-config` RPC definition's input container, which holds
the argument envelopes for the RPC.

Find was not finding these; while it validated the prefix and
traversed "ietf-netconf" module root to the "get-config" schema node
entry, it did not find "input", because the "input" and "output" path
elements are not inserted into the (*yang.Entry).Dir map.

Also link from the Input and Output entries to the parent (action or
RPC) entry so Path() works, among other things.

Add tests which will fail without the associated patches.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -647,7 +647,9 @@ func ToEntry(n Node) (e *Entry) {
 				if e.RPC == nil {
 					e.RPC = &RPCEntry{}
 				}
-				e.RPC.Input = ToEntry(i)
+				in := ToEntry(i)
+				in.Parent = e
+				e.RPC.Input = in
 				e.RPC.Input.Name = "input"
 				e.RPC.Input.Kind = InputEntry
 			}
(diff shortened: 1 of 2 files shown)


pkg/yang/entry.go
```diff
@@ -647,7 +647,9 @@ func ToEntry(n Node) (e *Entry) {
 				if e.RPC == nil {
 					e.RPC = &RPCEntry{}
 				}
-				e.RPC.Input = ToEntry(i)
+				in := ToEntry(i)
+				in.Parent = e
+				e.RPC.Input = in
```
andaru: Alternative: I wasn't as sure about directly assigning 
```
  e.Dir["input"] = in
```
at this point; that would avoid the additional `e.RPC != nil` case during `Find`, below by just making `e.Dir` track the schema node identifiers in this case; `input` and `output` are valid `schema-node-identifier` elements.

Thoughts?  If we're meant to see child identifiers as keys in `Dir`, it seems correct to add the `'input` or `'output` key to `e.Dir` in the code block here.

pborman: I think this should work.  It is a breaking change since the RPC field of Entry would go away.  It is not clear if we would need to add an RPCEntry kind to distinguish RPC entries from other containers.  I will approve this change as it is a non-breaking change.  Using Dir instead of RPC can be decided upon after this goes in.

andaru: Sounds good, thanks for the comments.

PR #70, opened 2017-12-22
Set leaf-list and leaf prefixes

rski:


andaru:
Looks reasonable from here; @robshakir, @pborman - any comments?

robshakir:
LGTM!

PR #78, opened 2018-04-12
Allow "description" and "reference" inside "import".

### LINKED ISSUE #76: "reference" and "description" inside "import" is not supported
YANG 1.1 allows usage of "reference" and "description" statements inside "import" statement.

Example of such usage is in [ietf-routing](https://tools.ietf.org/html/rfc8349#section-7) from [RFC 8349](https://tools.ietf.org/html/rfc8349).

idryzhov:
Fixes #76.

### DIFF
--- a/pkg/yang/yang.go
+++ b/pkg/yang/yang.go
@@ -210,7 +210,9 @@ type Import struct {
 	Parent     Node         `yang:"Parent,nomerge"`
 	Extensions []*Statement `yang:"Ext"`
 
+	Description  *Value `yang:"description"`
 	Prefix       *Value `yang:"prefix,required"`
+	Reference    *Value `yang:"reference"`
 	RevisionDate *Value `yang:"revision-date"`
 
 	// Module is the imported module.  The types and groupings are


aashaikh:
I do have a bit of concern about the mixing of 1.0 and 1.1 validation here.   For `when` statements, the RFCs seems vague on what the allowed substatements are, so it's seems ok that we might selectively support some.  For `import`, however, the spec for 6020 and 7950 are clear.  Would we position this as the beginning of 1.1 support in goyang?  :-)

robshakir:
I think we should merge this -- I have significantly less concerns about mixing of validation than @aashaikh. `pkg/yang` in general is permissive, it doesn't try and validate anything about the YANG that it produces output for. We also already have support for Y1.1 statements that would not be gated on YANG version (e.g., `action`).

I think that running the module through `pyang` or `yanglint` is likely a good first step for the `goyang` user, and downstream code consumers (e.g., `ygot`) which have significantly more knowledge of what they can produce from the input YANG can be more closed if they like.

@aashaikh PTAL in the light of the above, I +1 with @andaru to merge this.

guilt:
Yeah, let's get these changes merged.

LeonGWang:
I second this. Let's merge this.

robshakir:
I think we should merge #92, since it has testing -- I'd forgotten this PR, apologies.

idryzhov:
Implemented in #92

PR #81, opened 2018-11-02
Change FixChoice()'s inserted entry's Node type to Case

LeonGWang:
The inserted case entry's Node type should be the same as a real case entry.


robshakir:
This looks good -- could you add a test that covers this issue?  It looks like we have relatively minimal coverage for `FixChoice` and it won't be well tested elsewhere in the OC code, since OpenConfig does not use `choice`.

LeonGWang:
Sorry for the delay. I have added the test case for FixChoice(). Let me know if this suffices.

andaru:
LGTM, thanks Leon!

PR #82, opened 2018-11-10
Add initial support for deviations.

robshakir:
This change is to be staged into a number of sub-changes to be
reviewed -- it adds support for the deviation statement to goyang.

@aashaikh -- it'd be good to discuss this one.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -71,10 +71,12 @@ type Entry struct {
 	Name        string    // our name, same as the key in our parent Dirs
 	Description string    `json:",omitempty"` // description from node, if any
 	Default     string    `json:",omitempty"` // default from node, if any
-	Errors      []error   `json:"-"`          // list of errors encountered on this node
+	Units       string    `json:",omitempty"`
+	Errors      []error   `json:"-"` // list of errors encountered on this node
 	Kind        EntryKind // kind of Entry
 	Config      TriState  // config state of this entry, if known
 	Prefix      *Value    `json:",omitempty"` // prefix to use from this point down
+	Mandatory   TriState  `json:",omitempty"` // whether this entry is mandatory in the tree
 
 	// Fields associated with directory nodes
 	Dir map[string]*Entry `json:",omitempty"`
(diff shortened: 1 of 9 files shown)


pkg/yang/entry.go
```diff
@@ -675,34 +727,120 @@ func ToEntry(n Node) (e *Entry) {
...
+			// The type keyword is specific to deviate to change a type. Other type handling
+			// (e.g., leaf type resolution) is done outside of this case.
+			n, ok := n.(*Deviate)
+			if !ok {
+				e.addError(fmt.Errorf("unexpected type found, only valid under Deviate, is %T", n))
+				continue
+			}
 
+			if n.Type != nil {
+				if errs := n.Type.resolve(); errs != nil {
```
andaru: overly cautious question: could we refer to a type in an earlier `deviate` statement, which we then, in a later deviation, set via another deviate stmt to `not-supported`?

pkg/yang/entry.go
```diff
@@ -675,34 +727,120 @@ func ToEntry(n Node) (e *Entry) {
...
 						}
 					}
 				}
 			}
+		case "deviate":
+			if a := fv.Interface().([]*Deviate); a != nil {
+				for _, d := range a {
+					de := ToEntry(d)
+					var arg deviationType
+					switch d.Statement().Argument {
```
andaru: curious: Is there any preference between using a map and a choice statement? (the other way around in the String() function uses a `map`)

pkg/yang/entry.go
```diff
@@ -675,34 +727,120 @@ func ToEntry(n Node) (e *Entry) {
...
+						arg = DeviationReplace
+					case "delete":
+						arg = DeviationDelete
+					}
+
+					if e.Deviate == nil {
+						e.Deviate = map[deviationType][]*Entry{}
+					}
+
+					if _, ok := e.Deviate[arg]; !ok {
```
andaru: Since the map is of value types, I believe you can avoid the existence check and slice creation (796-798) entirely, because the `[]*Entry{}(nil)`  you get back from the `e.Deviate[arg]` map lookup on line 799 is valid to be appended to.

pkg/yang/entry.go
robshakir: Either way, it should be consistent! I don't think that there's a preferred type for me, or style here. I've adopted a `map` consistently here.

pkg/yang/entry.go
robshakir: Thanks! Fixed.

PR #83, opened 2018-11-12
Remove protobuf rendering, replaced by ygot.

robshakir:
The current `proto` output in this repo is incomplete, and ygot has implemented something more fully featured. This PR proposes removing the the `proto` output from goyang.


robshakir:
Merging this based on no objections being raised.

PR #84, opened 2018-11-23
fix link in README.md

qin-nz:
origin `golang.org/dl` is link to `https://github.com/openconfig/goyang/blob/master/golang.org/dl`


andaru:
Thank you for the correction!

PR #86, opened 2018-11-26
entry: support "rpc" with no "input" or "output" child

### LINKED ISSUE #85: Support "rpc" statements without "input" or "output" child statements
Presently, an RPC statement without a child "input" or "output" statement (valid per RFC) produces a `*yang.Entry` with a nil `RPC` field after a call to `ToEntry`.

This makes such RPC difficult to use with goyang.

Instead, I would expect to see a populated `(*yang.Entry).RPC{}` (type `*yang.RPCEntry`) whose `Input` and `Output` fields are nil.

andaru:
An "rpc" statement can have no "input" or "output" children according
to both RFC6020 and RFC7950.  Presently, when goyang encounters
neither of these children, it will not set (*yang.Entry).RPC to a
non-nil value, making these types of RPCs difficult to use with
goyang.

Examples of such RPCs: ietf-netconf.yang's "close-session" and
"discard-changes".

Resolves #85.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -705,7 +705,14 @@ func ToEntry(n Node) (e *Entry) {
 			// seems fine to ignore them for now, we are
 			// just interested in the tree structure.
 			for _, r := range fv.Interface().([]*RPC) {
-				e.add(r.Name, ToEntry(r))
+				switch rpc := ToEntry(r); {
+				case rpc.RPC == nil:
+					// When "rpc" has no "input" or "output" children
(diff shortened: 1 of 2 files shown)


andaru:
Aborting; will repost.

andaru:
This is exactly what I was planning to post; re-opening.

andaru:
Folks, I'd like to merge this in the next day; if you have comments, please post them.

andaru:
Cheers, Rob, will do!  Thanks, -a

PR #88, opened 2018-12-03
Resolve duplicate deviation processing with non-nil revision stmt.

robshakir:
This PR resolves an issue whereby if a module specified a revision
statement and contained deviations then the deviations would be
processed for both modname and modname@revision such that not-supported
deviations then could not be found.


robshakir:
Merging this to fix HEAD.

PR #92, opened 2019-02-25
Add support for the 'reference' statement under import.

robshakir:
This change adds support for the YANG 1.1 construct:

```yang
import foo {
  prefix "bar";
  reference "bat";
}
```

to goyang. The test added provides an extensible way to add tests for the contents of opaque `Node` structures to the goyang pkg.

PR #93, opened 2019-03-12
Add support for description and reference in import block.

guilt:
We require this one for the new Yang models: yang-schema-mount and yang-network-instance

based on Yang 1.1


robshakir:
This looks similar as a change to that which I made in #92 -- except you also added `Description`. I prefer the testing approach here -- since rather than an error, we check that we got the node that we expected. You can either cherry-pick my code from there, or I'll merge `Description` into that PR. Up to you :-)

Thanks for the change!
r.

guilt:
@robshakir Do add a Description in your PR and let's get it in! :)

guilt:
Already merged. Closing as duplicate.

Issue #51, opened 2017-06-26
Supporting YANG 1.1 "uses": changes to "augment" cardinality


andaru:
In evaluating what changes are necessary to support YANG 1.1 completely in the lexer, I've found one breaking change;

In YANG 1.0, a `uses` statement has a child statement `augment`, with a cardinality of `0..1` (ref: [RFC6020 s7.12.1](https://tools.ietf.org/html/rfc6020#section-7.12.1)).

In YANG 1.1, a `uses` statement has a child statement `augment`, with a cardinality of `0..n` (ref: [RFC7951 s7.13.1](https://tools.ietf.org/html/rfc7950#section-7.13.1)).

Due to the use of reflection in the YANG lexer, to support this change we could change one field in the `Uses` type;

```
   Augment *Augment `yang:"augment"`
```
to:
```
   Augment []*Augment `yang:"augment"`
```

However, this causes a breaking change for library users.  The `Augment` field of `Uses` is not used by the `goyang` library itself at this point.

Multiple `augment` children of a `uses` is found in both `ietf-ipv4-unicast-routing@2016-11-04.yang` and `ietf-ipv6-unicast-routing@2016-11-04.yang`, e.g.,

https://github.com/YangModels/yang/blob/master/standard/ietf/RFC/ietf-ipv4-unicast-routing%402016-11-04.yang#L222-L244

Thoughts on the most appropriate way to make this change?

pborman:
I think it would be appropriate to have a 1.0 and 1.1 version of goyang.  Probably create pkg/yang/v1.1, clone the source code there, and keep it with the package name of yang.  The import line would then be:

import "github.com/openconfig/goyang/pkg/yang/v1.1"

And then mark that package as in flux for now (i.e., we will make breaking changes to migrate from 1.0 to 1.1)

What do you think?

andaru:
Paul, thanks for the response.  Sounds like a fine approach, I'll start a new branch with those changes.

While I guess we haven't reached a "stable" version yet, have you had any thoughts about use of `gopkg.in` for publicly versioning the package?

LeonGWang:
Hi, I would like to help resolve this issue.
@robshakir do you think we should version `goyang`? If so, how do you envision such change?

robshakir:
I think there are a couple of options.

1. We can fork the package like @pborman suggested above. Downside for this is that we need to backport fixes across the board.
[...]

PR #94, opened 2019-03-12
Update `Uses` to work with Yang 1.1 files.

guilt:
Yang 1.1 files use multiple Augments in them.

### DIFF
--- a/pkg/yang/yang.go
+++ b/pkg/yang/yang.go
@@ -626,13 +626,13 @@ type Uses struct {
 	Parent     Node         `yang:"Parent,nomerge" json:"-"`
 	Extensions []*Statement `yang:"Ext" json:"-"`
 
-	Augment     *Augment  `yang:"augment" json:",omitempty"`
-	Description *Value    `yang:"description" json:",omitempty"`
-	IfFeature   []*Value  `yang:"if-feature" json:"-"`
-	Refine      []*Refine `yang:"refine" json:"-"`
-	Reference   *Value    `yang:"reference" json:"-"`
-	Status      *Value    `yang:"status" json:"-"`
-	When        *Value    `yang:"when" json:",omitempty"`
+	Augment     []*Augment `yang:"augment" json:"-"`
+	Description *Value     `yang:"description" json:",omitempty"`
+	IfFeature   []*Value   `yang:"if-feature" json:"-"`
+	Refine      []*Refine  `yang:"refine" json:"-"`
+	Reference   *Value     `yang:"reference" json:"-"`
+	Status      *Value     `yang:"status" json:"-"`
+	When        *Value     `yang:"when" json:",omitempty"`
 }
 
 func (Uses) Kind() string             { return "uses" }


LeonGWang:
This PR might not be so straightforward. Issue #51 discusses this in more detail.

pkg/yang/yang.go
```diff
@@ -626,13 +626,13 @@ type Uses struct {
...
 	Extensions []*Statement `yang:"Ext" json:"-"`
 
-	Augment     *Augment  `yang:"augment" json:",omitempty"`
-	Description *Value    `yang:"description" json:",omitempty"`
-	IfFeature   []*Value  `yang:"if-feature" json:"-"`
-	Refine      []*Refine `yang:"refine" json:"-"`
-	Reference   *Value    `yang:"reference" json:"-"`
-	Status      *Value    `yang:"status" json:"-"`
-	When        *Value    `yang:"when" json:",omitempty"`
+	Augment     []*Augment `yang:"augment" json:"-"`
```
andaru: This change to an exported field makes your PR not-backwards compatible with folks currently using this field with previous `goyang` releases - we'd prefer to avoid this outcome.

What we'd discussed (sorry, the PR/issue number with the discussion escapes me right now) was to instead have this:

```go
  Augment *Augment `yang:"augment" ...` // for 0th "augment"
  Augments []*Augment `...` // for 1..Nth "augment"
```

Adding the s-suffixed field, implying plurality, stores only the 1..Nth such statement, facilitating backwards compatible use.

To support the above will require some changes to the tag handling code (e.g., here https://github.com/openconfig/goyang/blob/b6e1e28273ad274f9240361b3a1ae65d27c045ac/pkg/yang/node.go#L245), e.g., using a whitelist of field names which are "plural" fields, or changes to the tag structure itself.  The former of those (statically describing plural field literals), though not ideal, seems preferable to me.  My reasoning is that tag reflection has some magic (until you're familiar with both how struct tagging and reflection works and read the specific code in goyang which deals with its tags), while a `switch` statement for the static plural cases is obvious.

Any chance you could give the above approach a try and see if there's any big roadblocks?

andaru: Oh, thanks @LeonGWang (seeing your comment from three weeks back) - that is the issue in question.

guilt: @andaru I see, but I had originally pushed this because I wanted the Yang to parse, whereas the codegen with the parsed Yang model was not using these Augments at all (within the Uses block) - was a separate issue to fix.

Could we instead opt to keep Augment == Augments[0] ? that way, you could do both, and going through augments does not involve a custom iteration / chain.

andaru: There is not any use of the 0, 1..N recipe presently.

Yes, we could have the first `augment` be stored in both `Augment` and `Augments[0]`.  That maintains backward compatibility perfectly.  While it may not be a concern of some users, some users are concerned about memory consumption (see PR 95 for example), so I didn't originally suggest it, but I think it's reasonable.

A third approach (briefly mentioned towards the end of issue 51) would avoid the duplicate storage by adding a method to (each) type.

```go
func (o *Uses) GetAugment() []*Augment { ... }
```

This does (as I understand it) add to the static object/binary size, however, so it's not a clear winner either.

Any comments (from yourself or others, eg @LeonGWang, @robshakir ?) regarding the above?

andaru: It looks like a change got merged which used `Augment []*Augment` instead of separating the fields out.  Would you mind closing this PR or splitting out any other work you needed into this or a separate PR?

guilt: I'm glad this is resolved now. :-) Thank you for the support and getting the Yang functionality in.

@andaru @LeonGWang

guilt:
Closing pull request. Thanks!

PR #95, opened 2019-04-04
Flag-control whether new 'Uses' field is populated.

robshakir:
```
 * (M) pkg/yang/{entry,entry_test,options}.go
  - Control whether Uses is populated based on an input flag. With
    large sets of schemas, the Uses data can cause a significant increase
    in the output size of the Entry data - which has memory footprint
    implications. Flag control the population of such fields.
```


robshakir:
@LeonGWang PTAL -- we ran into some performance impact internally when adopting this change. I propose to make it flag-controlled.

LeonGWang:
@robshakir LGTM. Let's merge it!

andaru:
@robshakir LGTM, thanks.

all: What was the original problem which led to these `uses` statements being tracked?   Was the problem related to identifier namespaces or other identifier resolution issues?  

If so, it should be that we can avoid back-propagating the pointers to the `uses` statements by tracking those in the "forward" direction as we traverse the populated model.

robshakir:
The issue is supporting the `uses .... { when ... } ` and `augment ... { when ... } ` statements. When we convert to entries today, we lose any context as to which `grouping` or `augment` statement contained any entry to start with. 

An alternative to tracking this per parent `Entry` (what was implemented here) is for us to rather track these in the child entries -- however, this would then be separate to where we store the contents of the `when` statement. We essentially would need some new way to describe these conditions - and then subsequently validate them.

robshakir:
Thanks for the review -- I'll merge here, and then open a new issue for us to discuss this design.

PR #98, opened 2019-09-20
Fix RFC references in doc comments.

nkitchen:
References to RFC 7950 were copied with correct numbers but all noted as '"action" sub-statement'.


LeonGWang:
LGTM. Let's merge this.

PR #101, opened 2019-10-17
Add support for "if-feature" statements as a child of "bit", "enum", …

darwish:
…"identity", and "refine"

See https://tools.ietf.org/html/rfc7950#section-1.1


darwish:
@googlebot I signed it!

darwish:
@googlebot I signed it!

darwish:
@googlebot I fixed it.

PR #102, opened 2019-10-18
Add support for "if-feature" statements as a child of "bit", "enum", …

midakaloom:
…"identity", and "refine"

Also add support for notifications to be tied to data nodes.

See https://tools.ietf.org/html/rfc7950#section-1.1

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -548,7 +548,7 @@ func ToEntry(n Node) (e *Entry) {
 		e.Config, err = tristateValue(s.Config)
 		e.addError(err)
 		e.Prefix = getRootPrefix(e)
-		addExtrasToLeaf(n, e)
+		addExtraKeywordsToLeafEntry(n, e)
 		return e
 	case *LeafList:
 		// Create the equivalent leaf element that we are a list of.
@@ -576,7 +576,7 @@ func ToEntry(n Node) (e *Entry) {
 			OrderedBy:   s.OrderedBy,
 		}
 		e.Prefix = getRootPrefix(e)
-		addExtrasToLeaf(n, e)
+		addExtraKeywordsToLeafEntry(n, e)
 		return e
 	case *Uses:
 		g := FindGrouping(s, s.Name, map[string]bool{})
@@ -587,7 +587,7 @@ func ToEntry(n Node) (e *Entry) {
 		// when the group is used in multiple locations and the
 		// grouping has a leafref that references outside the group.
 		e := ToEntry(g).dup()
-		addExtrasToLeaf(n, e)
+		addExtraKeywordsToLeafEntry(n, e)
 		return e
 	}
 
@@ -947,8 +947,8 @@ func ToEntry(n Node) (e *Entry) {
 	return e
 }
 
-// addExtrasToLeaf stores the values for unimplemented keywords in leaf entries.
-func addExtrasToLeaf(n Node, e *Entry) {
+// addExtraKeywordsToLeafEntry stores the values for unimplemented keywords in leaf entries.
+func addExtraKeywordsToLeafEntry(n Node, e *Entry) {
 	v := reflect.ValueOf(n).Elem()
 	t := v.Type()
 
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -548,6 +548,7 @@ func ToEntry(n Node) (e *Entry) {
 		e.Config, err = tristateValue(s.Config)
 		e.addError(err)
 		e.Prefix = getRootPrefix(e)
+		addExtrasToLeaf(n, e)
 		return e
 	case *LeafList:
 		// Create the equivalent leaf element that we are a list of.
@@ -575,6 +576,7 @@ func ToEntry(n Node) (e *Entry) {
 			OrderedBy:   s.OrderedBy,
 		}
 		e.Prefix = getRootPrefix(e)
+		addExtrasToLeaf(n, e)
 		return e
 	case *Uses:
 		g := FindGrouping(s, s.Name, map[string]bool{})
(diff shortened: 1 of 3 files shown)


midakaloom:
@googlebot I signed it!

midakaloom:
@robshakir I've added some tests

midakaloom:
@robshakir Any update on this?

pkg/yang/entry_test.go
```diff
@@ -1658,6 +1658,310 @@ func TestActionRPC(t *testing.T) {
...
+	in   string
+}{
+	{
+		name: "if-feature.yang",
+		in: `module if-feature {
+  namespace "urn:if-feature";
+  prefix "feat";
+
+  container cont {
+    if-feature ft;
```
andaru, line 1672: Would you mind adding/changing these feature names to match their usage, so that the same feature label/name is used only for one test assertion?

e.g., for this case: `if-feature ft-container;`

pkg/yang/entry_test.go
```diff
@@ -1658,6 +1658,310 @@ func TestActionRPC(t *testing.T) {
...
+    action act {
+      if-feature ft;
```
andaru, line 1674: As above; e.g., 

`if-feature ft-action;`

pkg/yang/entry_test.go
```diff
@@ -1658,6 +1658,310 @@ func TestActionRPC(t *testing.T) {
...
+    }
+  }
+
+  anydata data {
+    if-feature ft1;
```
andaru, line 1679: These names are probably ok since they're uniquely used on this `anydata` node.

pkg/yang/entry_test.go
```diff
@@ -1658,6 +1658,310 @@ func TestActionRPC(t *testing.T) {
...
+    if-feature ft2;
+  }
+
+  anyxml xml {
+    if-feature ft;
```
andaru, line 1684: Rename feature as above (and other instances of feature name `ft` below).

pkg/yang/entry_test.go
```diff
@@ -1658,6 +1658,310 @@ func TestActionRPC(t *testing.T) {
...
+	testcases := []struct {
+		name           string
+		inIfFeatures   []*Value
+		wantIfFeatures []string
+	}{
+		// Node statements
+		{
+			name:           "action",
+			inIfFeatures:   entryIfFeatures(mod.Dir["cont"].Dir["act"]),
+			wantIfFeatures: []string{"ft"},
```
andaru, line 1779: `ft-action` and etc below.

Yes, this really just makes it easier for reviewers to ensure that the test cases are matching specific test data without checking the code and attempting test breakage to ensure the appropriate labels are being referred to!  (Much appreciated)

pkg/yang/yang.go
```diff
@@ -339,25 +339,26 @@ type Container struct {
...
+	Config       *Value          `yang:"config"`
+	Container    []*Container    `yang:"container"`
+	Description  *Value          `yang:"description"`
+	Grouping     []*Grouping     `yang:"grouping"`
+	IfFeature    []*Value        `yang:"if-feature"`
+	Leaf         []*Leaf         `yang:"leaf"`
+	LeafList     []*LeafList     `yang:"leaf-list"`
+	List         []*List         `yang:"list"`
+	Must         []*Must         `yang:"must"`
+	Notification []*Notification `yang:"notification"`
```
andaru, line 355: Just confirming, did these `notification` lines come in from rebasing some other commits?

pkg/yang/entry.go
```diff
@@ -943,6 +947,30 @@ func ToEntry(n Node) (e *Entry) {
 	return e
 }
 
+// addExtrasToLeaf stores the values for unimplemented keywords in leaf entries.
+func addExtrasToLeaf(n Node, e *Entry) {
```
andaru, line 951: Slightly pedantically, perhaps `addExtraKeywordsToLeafEntry`?  

Mostly to avoid someone coming along later and saying "oh, I can add more extra stuff here", when this mightn't be the right place for another use case.

andaru:
I notice in the tests that the feature identifier names aren't defined.  

While your change doesn't strictly need that, ideally the parser should (after all files have been loaded) produce errors for labels which are used and which cannot be resolved.  

Do you think it should do that? (i.e., once all files are loaded as the identifiers may be defined in an included submodule).

pkg/yang/yang.go
midakaloom: No, it came from me implementing a YANG module that defined notifications not at the top-level, and realizing that support was lacking here.

I can split it out into a separate PR if you prefer.

pkg/yang/entry.go
midakaloom: Good call. Done.

pkg/yang/entry_test.go
midakaloom: Done, along with all other similar comments.

midakaloom:
> I notice in the tests that the feature identifier names aren't defined.
> 
> While your change doesn't strictly need that, ideally the parser should (after all files have been loaded) produce errors for labels which are used and which cannot be resolved.
> 
> Do you think it should do that? (i.e., once all files are loaded as the identifiers may be defined in an included submodule). [edit: I'm not suggesting you do that here - if your reading of the specification leads you to think the parser should identify missing features, I'll open a separate issue for us to address].

I think you're right, and according to https://tools.ietf.org/html/rfc7950#section-7.20.1 the parser should identify the missing features. This also matches the behaviour I see in the 3rd party netconf server I tried it with.

For now, I've added the feature definitions to the test module for correctness.

hellt:
Hi all,
is anything else needed for that PR to be merged?

hellt:
Hi @andaru @robshakir @midakaloom 
what comments are left unaddressed? I would like to help since this _feature_ is needed.

andaru:
No, I think everything's been addressed.  Thank you!

If there's further follow up work re: Notifications, would you open an issue with those details if not already, please?

andaru:
Rob's requested change about updating tests looks addressed to me in terms of what he mentioned, and I'm happy with it, so unless I hear otherwise in the next few hours, I'll merge it.  Sound good?

midakaloom:
Sounds good to me, thanks.

robshakir:
@andaru - thanks for shepherding the changes here. Feel free to go ahead and merge this. Thanks!

Issue #103, opened 2019-10-22
Memory issue due to github.com/openconfig/goyang/pkg/yang#typeDictionary


ivajloip:
Hello,

I am using your great library while working on an implementation of a library that is parsing yang files and does some processing with them (in relation to [this][1]). Recently I realized parsing many models makes my library leak memory. After some debugging I discovered that it should be because of the use of [typeDict][2] that is growing with each model parsing. I would be willing to work on resolving this issue, but as I saw that there is already a `TODO` comment around it, I imagine you already have an idea what would be a good solution for you. The best solution for me would be to have reference to the `typeDict` in the `Modules` type, but it seems that some initialization is done for the _default_ types, which would make that more difficult (we can have separate `typeDict` for the built-in types and the parsed ones). Do you have better ideas and would you be willing to review and possibly merge such a change?

Thank you in advance!

[1]: https://datatracker.ietf.org/doc/draft-petrov-t2trg-youpi/
[2]: https://github.com/openconfig/goyang/blob/master/pkg/yang/types.go#L41

robshakir:
Hi Ivaylo,

The `TODO` here was Paul's initial thoughts about how this might change, but I don't think it's been a problem for us, such that we haven't placed much priority on amending this TODO.  Our applications, today, usually parse the YANG modules and then output some other code that we use elsewhere (e.g., [ygot](https://github.com/openconfig/ygot). I can see that if you're doing a lot of module processing, then this would grow without bounds (since it's not cleaned up).

I'm open to a change whereby we make this part of the `Modules` type - such that within a particular context then this would be cleaned up when that parsing is discarded. I'm open to reviewing changes for this -- please feel free to open a PR.

Cheers,
r.

ivajloip:
Hi Rob,

Thank you for the quick reply and the expressed willingness to review a possible change. The proposed fix is now published. If you need clarifications or other changes, please don't hesitate to let me know. One Possible modification that I was considering, but decided to wait for your input before implementing it, is to try to add some tests related to the fixed issue. My concern about this is that the test that I could think of, which is quite similar to what I did to test the proposed change locally, is to run a good number of parsings and observe if the application has allocated more than some amount of memory. In order for the tests to be more reliable, the number of parsings need to be high enough, which might take at least few seconds, possibly more depending on the hardware running the tests. I am not sure if such test/s would be something you would want to have.

[...]

PR #104, opened 2019-10-23
Fix memory leak when parsing multiple modules

ivajloip:
Resolves issue #103


ivajloip:
@googlebot I signed it!

pkg/yang/modules.go
```diff
@@ -28,16 +28,27 @@ type Modules struct {
...
 // NewModules returns a newly created and initialized Modules.
 func NewModules() *Modules {
+	d := &typeDictionary{dict: map[Node]map[string]*Typedef{}}
+	// taking the lock should not be needed as this type should only be modified
+	// during init.
+	typeDict.mu.Lock()
+	defer typeDict.mu.Unlock()
+	for k, v := range typeDict.dict {
+		d.dict[k] = v
+	}
```
andaru, line 43: I wonder if it might be less confusing to have a package default `Modules` variable, rather than a package default `typeDict` as we have now, and a separate copy in each instance of `Modules`?

This way, every time a reference is made to a `typeDict`, it is being made to a specific `(*Modules).typeDict`.

e.g.,
```go
func init() {
  defaultModules = NewModules()
}
```

and `NewModules` would include the initTypes call to the top level statement, e.g., something like

```go
initTypes(reflect.TypeOf(&meta{}), xyz)
```

Though, this would mean refactoring much of your change to refer to a `Modules` rather than a `typeDict`...

Given this might change the overall structure of your change, I'll leave the review here for the moment to get your input and the other maintainers' input on the factoring.

nadddy:
Is there a plan to merge this PR. We have a use-case of large yang modules which are dynamically loaded and unloaded. Because of this global variable we never get memory freed up after modules are removed.

nadddy:
@robshakir Is there a concern with this PR? I can spend some time to refactor if current approach is not acceptable. Since this is related to memory leak please help resolve this issue.

pkg/yang/modules.go
wenovus: I think putting the global cache objects into a global/default `Modules` struct might not be necessary because I'm not sure it serves any purpose.

Once we manage to put all the global cache objects into the `Modules` struct, each `Modules` is going to be isolated from each other, as well as from the global/default cache objects. Now, these global/default cache objects will only be populated and accessed by lower-level calls (i.e. `Parse`, `BuildAST`, `ToEntry`), so even if we have a global/default `Modules` object, the user will never have access to it via any of these calls, and therefore, its fields, like `Modules`, `SubModules`, etc. will not be populated either. So I think just keeping the global/default cache objects might be enough in the longterm if we decide to keep the lower-level calls for backwards compatibility.

wenovus, line 43: I'm confused about this part -- wouldn't this be polluting the type space of the `Modules` object with whatever has been populated into the default/global `typeDict`, which I expect to be empty unless the user has decided to make lower-level calls (`Parse`, `BuildAST`)?

wenovus (commented):
I think I'm happy with the general direction of this change. Eventually we'd probably want all global caches to reside in the `Modules` objects. I think this might just be one place where memory is persisting across different parsing calls.

However I am confused by the copy from the global/default `typeDict` to the local `typeDict`, I'm wondering if you remember why this was needed?

wenovus:
I made some changes at 4dc31d2, PTAL.

wenovus:
@robshakir I used the [replace directive](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) to work around modules and test that ygot is unaffected by this change.

pkg/yang/ast.go
```diff
@@ -87,15 +88,15 @@ func BuildAST(s *Statement) (Node, error) {
 
 // build builds and returns an AST from the statement s, with parent p, or
 // returns an error.  The type of value returned depends on the keyword in s.
-func build(s *Statement, p reflect.Value) (v reflect.Value, err error) {
+func build(s *Statement, p reflect.Value, d *typeDictionary) (v reflect.Value, err error) {
```
robshakir, line 91: Update documentation comment?

pkg/yang/ast.go
```diff
@@ -78,7 +74,12 @@ var aliases = map[string]string{
 // BuildAST builds an abstract syntax tree based on the yang statement s.
 // Normally it should return a *Module.
 func BuildAST(s *Statement) (Node, error) {
-	v, err := build(s, nilValue)
+	return buildASTWithTypeDict(s, &typeDict)
+}
+
+func buildASTWithTypeDict(s *Statement, d *typeDictionary) (Node, error) {
```
robshakir, line 80: Documentation comment?

pkg/yang/ast.go
```diff
@@ -258,7 +259,7 @@ func build(s *Statement, p reflect.Value) (v reflect.Value, err error) {
 //                   (This is to support merging Module and SubModule).
 //
 // If at contains substructures, initTypes recurses on the substructures.
-func initTypes(at reflect.Type) {
+func initTypes(at reflect.Type, d *typeDictionary) {
```
robshakir, line 262: Documentation comment update?

robshakir (changes requested):
LGTM as a change -- just propose that we update the comments to point out the type dict is now handed around within a particular context.

For the ygot tests -- did you regenerate all the integration schemas?

wenovus:
> LGTM as a change -- just propose that we update the comments to point out the type dict is now handed around within a particular context.
> 
> For the ygot tests -- did you regenerate all the integration schemas?

Yes all the integration schemas pass.

robshakir (approved):

wenovus:
Thanks for the review!

wenovus (approved):
I'll also approve since Ivaylo is the original author and I just made some tweaks.

PR #109, opened 2019-12-18
Do not drop extensions when merging a grouping Entry

### LINKED ISSUE #108: 'Entry.Ext' field is lost during 'merge()'
@wenovus This is a correction to the misguided issue #106 

Given an example module like so, 
```
module example {
    yang-version "1";

    prefix "ex";
    namespace "https://example";

    import extensions { prefix exts; }

    container C {
        exts:extension 1;
        uses G {
            exts:extension 2;
        }
    }

    grouping G {
        exts:extension 3;
        
        leaf L {
            exts:extension 4;
            type int;
        } 

        container C2 {
            leaf L2 {
                type int;
            }
        }
    }
}
```
I would expect `ext:extension2` and `ext:extension3` are kept in the `Entry.Ext` field of all direct child `Entry` of `grouping G`. In another word, the above module should result in the same `Entry` tree as this module:
```
module example {
    yang-version "1";

    prefix "ex";
    namespace "https://example";

    import extensions { prefix exts; }

    container C {
        exts:extension 1;

        leaf L {
            exts:extension 2;
            exts:extension 3;
            exts:extension 4;
            type int;
        }

        container C2 {
            exts:extension 2;
            exts:extension 3;
            leaf L2 {
                type int;
            }
        }
    }
}
```
However, because `Entry.Ext` field of the merged `Entry` is not kept during `merge()`, `exts:extension 2` and `ext:extension 3` are lost. The `Entry` tree we get from our first example module is currently equivalent to this last module:
```
module example {
    yang-version "1";

    prefix "ex";
    namespace "https://example";

    import extensions { prefix exts; }

    container C {
        exts:extension 1;

        leaf L {
            exts:extension 4;
            type int;
        }

        container C2 {
            leaf L2 {
                type int;
            }
        }
    }
}
```

wenovus:
Currently, when the `merge` function carries out the transplant of top-level nodes from a grouping to their target parent, extensions at the grouping level, which are stored at a temporary Entry to which the top-level nodes are attached before transplanting, are being dropped. This is because the `merge` code forgets about the `Exts` field when it carries out the transplant operation.

This change simply transplants the extensions as well to all the top-level nodes to make the transplant more complete.

See example #108 

Fixes #108

### DIFF
--- a/pkg/yang/entry_test.go
+++ b/pkg/yang/entry_test.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/openconfig/gnmi/errdiff"
 )
 
@@ -1083,6 +1085,7 @@ func TestFullModuleProcess(t *testing.T) {
 		inModules        map[string]string
 		inIgnoreCircDeps bool
 		wantLeaves       map[string][]string
+		customVerify     func(t *testing.T, module *Entry)
 		wantErr          bool
 	}{{
 		name: "circular import via child",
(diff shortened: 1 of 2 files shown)


LeonGWang:
LGTM. Thanks Wen!

pkg/yang/entry_test.go
```diff
@@ -1297,6 +1300,112 @@ func TestFullModuleProcess(t *testing.T) {
...
+					ext:a-define "hello";
+					uses daughter-group {
+						ext:d-define "universe";
+					}
+				}
+
+				grouping daughter-group {
+					ext:b-define "world";
+
+					leaf l {
```
aashaikh: can we name these nodes in a way that makes it easier to understand what the tests should expect?  just would be a bit easier to read the tests below IMO.

wenovus: Done

LeonGWang:
When can we expect this PR to be accepted?

PR #115, opened 2020-03-19
Add support for coalescing decimal64 ranges.

robshakir:
``` 
* (M) pkg/yang/types_builtin.go
  - Add support for add() on a Number that represents as decimal64,
    rather than an integer number. This was required to support modules
    that have multiple ranges that need to be coalesced together. See
    github.com/openconfig/ygot/issues/361.
 * (R) pkg/yang/range_test.go -> pkg/yang/number_test.go
  - Rename range_test to number_test to make it more general.
  - Add unit testing for the .add() function that was previously not
    explicitly tested (other than during range testing).
```


pkg/yang/number_test.go
```diff
@@ -15,322 +15,338 @@
...
-	if e == nil {
-		return ""
+func TestCoalesce(t *testing.T) {
+	for x, tt := range []struct {
+		in, out YangRange
+	}{
+		{},
+		{YangRange{R(1, 4)}, YangRange{R(1, 4)}},
+		{YangRange{R(1, 2), R(3, 4)}, YangRange{R(1, 4)}},
+		{YangRange{R(1, 2), R(2, 4)}, YangRange{R(1, 4)}},
```
wenovus: coalesce is still treating the range as an integer range when coalescing, so it's erroneously coalescing decimal ranges:

TestCoalesce:
```
                {
                        YangRange{YRange{Min: FromFloat(1), Max: FromFloat(2.5)}, YRange{Min: FromFloat(3), Max: FromFloat(4)}},
                        YangRange{YRange{Min: FromFloat(1), Max: FromFloat(2.5)}, YRange{Min: FromFloat(3), Max: FromFloat(4)}},
                },
```
number_test.go:187: #3: got 1.0..4.0, want 1.0..2.5|3.0..4.0

We can fix by either
1. support decimal range coalescing by adding the fractional digit instead of add(1).
2. not coalesce for decimal ranges.

We can do either by adding to `ParseRanges` a YANG typename parameter or have two `ParseRanges` for ints and `decimal64`. If the latter, then we can optionally change the current `TestCoalesce` to `TestParseRanges` as maybe that's the better test to have (or I'd be happy to do it in a follow-up PR).

pkg/yang/types_builtin.go
```diff
@@ -616,6 +619,7 @@ func DecimalValueFromString(numStr string, fracDigRequired int) (n Number, err e
 	if fracDigRequired < 0 {
 		fracDigRequired = fracDig
```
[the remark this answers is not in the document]
wenovus: This is assuming we're supporting negative inputs for automatically deducing fractional digits.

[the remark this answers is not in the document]
wenovus: Maybe this is inconsequential though, because the YangRanges we're generating currently don't store the user-input fractional digit.

pkg/yang/number_test.go
```diff
@@ -15,322 +15,338 @@
...
+		},
+		{
+			r1: YangRange{R(1024, 65535)},
+			r2: YangRange{R(-999999, 4096), R(5120, useMax)},
+			ok: false,
+		},
+		{
+			r1: YangRange{R(1024, 65535)},
+			r2: YangRange{R(useMin, 4096), R(5120, 999999)},
+			ok: false,
```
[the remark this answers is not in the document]
wenovus: sorry this one is fine.

pkg/yang/number_test.go
wenovus: decided to support decimal range coalescing by splitting ParseRanges into a variation for int and decimal.

pkg/yang/types_builtin.go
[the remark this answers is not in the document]
wenovus: Removed ability to automatically deduce fractional digit for a decimal string:
RFC7950:
```
The "fraction-digits" statement, which is a substatement to the
   "type" statement, MUST be present if the type is "decimal64".
```

pkg/yang/types_builtin.go
```diff
@@ -601,8 +604,8 @@ func ParseNumber(s string) (n Number, err error) {
 // precision of numStr is used to set the number of fractional digits.
 // numStr must conform to Section 9.3.4.
 func DecimalValueFromString(numStr string, fracDigRequired int) (n Number, err error) {
-	if fracDigRequired > int(MaxFractionDigits) {
-		return n, fmt.Errorf("too many fraction digits %d > max of %d", fracDigRequired, MaxFractionDigits)
+	if fracDigRequired > int(MaxFractionDigits) || fracDigRequired < 1 {
```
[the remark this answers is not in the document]
wenovus: removed support for this.

pkg/yang/number_test.go
```diff
@@ -15,322 +15,338 @@
...
+	}
+}
+
+func TestRangeContains(t *testing.T) {
+	for x, tt := range []struct {
+		r1, r2 YangRange
+		ok     bool
+	}{
+		{ok: true},
+		{r1: YangRange{R(1, 2)}, ok: true},
```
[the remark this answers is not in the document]
wenovus: I'm deciding to move this issue to a future commit, there is already enough going on in this one.

pkg/yang/types_builtin.go
```diff
@@ -563,10 +566,11 @@ func FromFloat(f float64) Number {
 	return n
 }
 
-// ParseNumber returns s as a Number.  Numbers may be represented in decimal,
+// ParseInt returns s as a Number with FractionDigits=0.
 // octal, or hexadecimal using the standard prefix notations (e.g., 0 and 0x)
-func ParseNumber(s string) (n Number, err error) {
```
robshakir: Can we not keep the public API here -- since `ParseNumber` returned `Number` then it seems like it should be compatible with both? Is the concern that we can't infer what the `fracDigitRequired` argument should be?

pkg/yang/types_builtin.go
```diff
@@ -696,12 +714,8 @@ func (n Number) Int() (int64, error) {
...
 }
 
-// add adds i to n without checking overflow.  We really only need to be
-// able to add 1 for our code. panics if n is a decimal.
-func (n Number) add(i uint64) Number {
-	if n.IsDecimal() {
-		panic("cannot call add() on decimal number " + n.String())
-	}
+// addQuantum adds the smallest quantum to n without checking overflow.
+func (n Number) addQuantum(i uint64) Number {
```
robshakir: What do you mean by quantum here? I'd read this as meaning that we add the smallest increment that is possible in the number (e.g., `fraction-digits` of 1 would mean that we add 0.1), but then the `uint64` argument doesn't make sense.

pkg/yang/types_builtin.go
```diff
@@ -790,16 +804,41 @@ func (r YRange) Valid() bool {
...
+// output range is sorted and coalesced.
+func ParseRangesInt(s string) (YangRange, error) {
+	return parseRanges(s, false, 0)
+}
+
+// ParseRangesDecimal parses s into a series of ranges. Each individual range is in s
+// is separated by the pipe character (|).  The min and max value of a range
+// are separated by "..".  An error is returned if the range is invalid. The
+// output range is sorted and coalesced.
+func ParseRangesDecimal(s string, fracDigRequired uint8) (YangRange, error) {
```
robshakir: I guess I'm not quite clear why we need `fracDigRequired` here. Could you explain?

Here's why I ask. If we have:

```yang
leaf foo {
  type decimal64 {
    fraction-digits 10;
    range "1...3.14159";
 }
```

then this should be valid input. In the ranges we only really care about the upper and lower bounds, which don't need to have fixed precisions. The usage example in RFC6020 shows this kind of input as an example ([here](https://tools.ietf.org/html/rfc6020#section-9.3.5)), so I figure we should support it. If we assume that we don't need `fraction-digits` specified here, then we can keep the public API of `ParseRanges`.

pkg/yang/types_builtin.go
```diff
@@ -932,38 +980,33 @@ func (r YangRange) Equal(q YangRange) bool {
 
 // Contains returns true if all possible values in s are also possible values
 // in r.  An empty range is assumed to be min..max.
+// TODO(wenbli): The current behaviour of this is confusing and unexplained by
+// comments: why does the presence of min/max always cause the range to be
```
robshakir: If there is `min...max` (i.e., minimum AND maximum) then that should mean everything is contained, right?

pkg/yang/types_builtin_test.go
```diff
@@ -0,0 +1,634 @@
...
+	}
+	if b == useMax {
+		n2 = maxNumber
+	}
+	return YRange{n1, n2}
+}
+
+func TestRangeEqual(t *testing.T) {
+	for x, tt := range []struct {
+		r1, r2 YangRange
```
robshakir: Whilst we're tweaking these, should we name them in a way that would be more typical? i.e., `inTestRange` and `inContainedRange`? We could also make `ok` -> `want` per usual conventions.

pkg/yang/types_builtin.go
wenovus: Yes exactly, `12.34` -- is that a valid decimal or should it fail? We don't know unless we know that the fractional digit is >= 2 and < 18: https://tools.ietf.org/html/rfc7950#section-9.3.4

pkg/yang/types_builtin.go
wenovus: `addQuantum`'s only purpose right now is to support coalescing by incrementing the max of the previous range to see whether it overlaps with the next range. 

The fractional digit
```
controls the size of the minimum difference between values of a
   decimal64 type by restricting the value space to numbers that are
   expressible as "i x 10^-n" where n is the fraction-digits argument.
```
So, coalescing decimals is akin to coalescing integers after multiplying by 10^n.

pkg/yang/types_builtin.go
wenovus: The problem is if you have
```
leaf foo {
  type decimal64 {
    fraction-digits 1;
    range "1...3.14159";
 }
```
Then it should fail, but the current code doesn't fail because it automatically deduces the fraction-digits and ignores the value in the YANG file.

pkg/yang/types_builtin.go
wenovus: the problem is 1..2 contains min..max, and that doesn't make sense to me:
```
		{
			r1: YangRange{R(1, 10)},
			r2: YangRange{R(useMin, useMax)},
			ok: true,
		},
```

Issue #116, opened 2020-04-18
error channel is not consumed


andaru:
Issue #71 identifies two problems.  The first, as that issue describes, is related to escape character handling.

The second, shown in the error message provided in #71 is a deadlock, which occurs due to the lexer's `items` channel overflowing.

This issue is opened to address that part of the problem.  

At present, errors can be sent to the `items` channel an arbitrary number of times between each call to `(*lexer).NextToken()`, which is the only place which receives from the `items` channel.  Each call to `NextToken` will remove at most one item.  When multiple errors are detected (in issue #71) within the same comment, only a single `NextToken` call will occur, allowing the `items` channel to overflow in short order.

andaru:
Diagnosis:

The stack path causing the issue in #71 comes down to the `lexQString` function's inner `for {}` loop, specifically this code.

```golang
if !l.inPattern {
	l.ErrorfAt(line, col, `invalid escape sequence: \`+string(c))
}
text = append(text, '\\')
```

After this executes, the `for` loop continues, which due to there being no goroutine consuming the `items` channel, allows for the channel to fill up and the deadlock to occur.

An elegant solution might be to start a goroutine to consume the channel prior to `lexQString` ever being called.

Item consumption occurs via calling `NextToken` manually.  This is presently called only from within `(*parser).next()`, itself only called from `nextStatement` which it called by itself recursively and ultimately from `Parse`.

The current control flow does not immediately lend itself to setting up a goroutine to consume the channel, but an elegant approach to that may appear given a little more consideration.

The simple approach of exiting `NextToken` as soon as any error is seen would work, but will make the parser strictly "first error" only, which is less than ideal considering multiple errors are supported today, including up to [size of items channel] consecutive errors between other tokens.

Given this issue has existed in the library since release, I'll keep working on a more elegant approach with a goroutine started by the call to `Parse`.

pborman:
You can put a select where the message is written and indicate a fatal error if the default case occurs (since it only happens after repeated failures). Just a thought.

andaru:
Thanks, that's a good idea to replace the deadlock with an explicit crash.  I'll do that here.

Do you think it's worth looking at cleaning this up?  We have a channel (looks like it comes from Rob's lexer/parser example, a pattern I quite like..), plus a stack of items on the receiver side. Since the channel send and receive calls happen in the same stack (frame), we could eliminate the channel itself as there are no goroutines to co-ordinate.  Or we could refactor the receiver to properly run in its own goroutine.

PR #117, opened 2020-04-19
panic if the lexer items channel (would) overflow

andaru:
To avoid a deadlock when sending a *token to the lexer's items
channel - which can occur if too many errors occur within a single
token - instead panic when the items channel fills up.

Non-error items (i.e., normal tokens) should not overflow the channel,
so use a different panic message (indicating a library bug) if this
occurs.

Avoids a deadlock and addresses #116 .

### DIFF
--- a/pkg/yang/lex.go
+++ b/pkg/yang/lex.go
@@ -191,12 +191,21 @@ func (l *lexer) emitText(c code, text string) {
 	if l.debug {
 		fmt.Fprintf(os.Stderr, "%v: %q\n", c, text)
 	}
-	l.items <- &token{
+	select {
+	case l.items <- &token{
 		code: c,
 		Text: text,
 		File: l.file,
 		Line: l.sline,
 		Col:  l.scol + 1,
+	}:
+	default:
+		if c == tError {
+			panic(tooMany)
+		}
+		// regular tokens should never over-fill the items channel,
+		// so reaching here implies a library bug
+		panic("goyang bug: lexer items channel not being consumed")
 	}
 	l.consume()
 }
--- a/pkg/yang/parse_test.go
+++ b/pkg/yang/parse_test.go
@@ -389,3 +389,29 @@ module base {
 		}
 	}
 }
+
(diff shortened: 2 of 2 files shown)


andaru:
Are we being strict about having to increase or maintain exact coverage for commits these days?  

While I'm confident the existing library code can't cause the code path which will trigger the "bug" case on the second panic (but I didn't want to add a potential case where the code could later silently skip a channel send),  I could refactor the test to send 4 or more non error items to the channel [channel size is statically initialized to 3] to cause that line to be executed.

Thoughts?

pborman:
Can you just ignore the send failure if c == tError?  Perhaps check to see if there have been any errors and if so, ignore this one as the parser will fail anyhow.  The channel should probably be large enough to handle the number of errors that cause the parser to bail plus enough to handle any legitimate messages,  or is this actually a loop that will never terminate?  If so, the panic should be caught higher up the stack, if at all possible.   (Is it already?  I have not looked at the code base for quite a few years)

andaru:
> Can you just ignore the send failure if c == tError?

I tried that first.  Because the cursor (`Location`) is updated during handling of the item, tests fail due to error messages no longer being accurate.

> The channel should probably be large enough to handle the number of errors that cause the parser to bail plus enough to handle any legitimate messages, or is this actually a loop that will never terminate?

Yes, the code here is attempting to continue after producing the error, so when you have enough errors within a single comment (for example) to fill the channel, this error occurs.  The test case here is a regression test for the corresponding issue.

> If so, the panic should be caught higher up the stack, if at all possible. (Is it already? I have not looked at the code base for quite a few years)

It's not caught, it causes a panic rather than a deadlock.

More could be done, e.g.,

1. Correcting cursor updates (so they occur, whether or not an item is produced).
2. Not using a channel at all - there is no other goroutine, so we don't need one.  The execution flow here suits not having a channel better than refactoring the code to have a goroutine consuming the queue.

I tried to do all of 2, but it would turns out to be a larger refactoring than this issue alone, and I wanted to at least turn this issue from being a deadlock into a crash.

PR #118, opened 2020-04-28
Add a GitHub action for running go build and testing.

robshakir:
Migrate testing from travis-ci to github actions.


.github/workflows/go.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    - name: Build
+      run: go build -v ./...
+
+    - name: Test
+      run: go test -v -coverprofile=profile.cov  ./... 
+    
+    - name: Race Test
+      run: go test -race ./...
+
```
wenovus, line 43: Add `go vet`?

.github/workflows/go.yml
```diff
@@ -0,0 +1,61 @@
...
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            dep ensure
+        fi
```
wenovus, line 33: https://github.com/golang/dep#dep `as of Go 1.11, the Go project has officially adopted a different approach, based on the concept of Modules`
Since we're using modules now, do we still need this?

wenovus:
Can also add badge to README status

robshakir:
Thanks for the review @wenovus !

PR #120, opened 2020-04-30
Add oc-versions format to goyang for displaying openconfig-versions

wenovus:


yang.go
```diff
@@ -56,10 +56,11 @@ import (
 // Each format must register a formatter with register.  The function f will
 // be called once with the set of yang Entry trees generated.
 type formatter struct {
-	name  string
-	f     func(io.Writer, []*yang.Entry)
-	help  string
-	flags *getopt.Set
+	name              string
+	f                 func(io.Writer, []*yang.Entry)
+	includeSubmodules bool
```
andaru, line 61: Please move the bool to be the last field of the struct (alignment)

versions.go
```diff
@@ -0,0 +1,67 @@
...
+			continue
+		}
+
+		fmt.Fprintf(w, "%s.yang:", m.Name)
+
+		for _, e := range m.Extensions {
+			keywordParts := strings.Split(e.Keyword, ":")
+			if len(keywordParts) != 2 {
+				// Unrecognized extension declaration
+				continue
```
andaru, line 52: Like on line 42, could you write a message to `os.Stderr` indicating what format must be used by the user? (help the user to understand their mistake)

versions.go
```diff
@@ -0,0 +1,67 @@
...
+				extMod := yang.FindModuleByPrefix(m, pfx)
+				if extMod == nil {
+					fmt.Fprintf(os.Stderr, "unable to find module using prefix %q from referencing module %q\n", pfx, m.Name)
+				} else if extMod.Name == "openconfig-extensions" {
+					fmt.Fprintf(w, " openconfig-version:%q", e.Argument)
+				}
+			}
+		}
+
+		fmt.Fprintf(w, "\n")
```
andaru, line 65: Please add a test for this, covering the following cases:

1. Errors (module unavailable, invalid format).
2. A yang file which has no extensions.
3. A yang file which has 1 extension.
4. A yang file which has more than 1 extension.
5. There's also the case where the extension is `openconfig-version`, but the extension module's name isn't `openconfig-extensions`.  The code says don't print anything, a test should assert that this is expected.

yang.go
```diff
@@ -202,6 +204,14 @@ Formats:
 			names = append(names, m.Name)
 		}
 	}
+	if f.includeSubmodules {
+		for _, m := range ms.SubModules {
+			if mods[m.Name] == nil {
```
andaru, line 209: can `m` ever be nil at this point?  If it could be, perhaps the branch should instead be:

```go
if m != nil && mods[m.Name] == nil {
  // existing branch body goes here
}
```

Reason being that if `m` were nil, the branch could be executed twice, leading to `names` having a duplicate in it.

(I think `m` probably cannot be nil; if you're sure that's the case, ignore this comment)

versions.go
wenovus: Thanks for the review Andrew! After talking to @robshakir I'm planning on moving this piece of code out of goyang into the private repo openconfig/models-ci to support CI of the openconfig models repo, as using the pkg/yang package directly is easier than parsing the output of this formatter.

wenovus:
The openconfig-versions output is intended for CI, and using the pkg/yang package directly seems to be better than parsing the output from this formatter.

PR #121, opened 2020-05-06
Move sort from YangRang Validate to parseRanges

harshilshah611:
Validate is also called from ygot/ytypes/int_type.go to validate YangRange. This creates race conditions.

### DIFF
--- a/pkg/yang/types_builtin.go
+++ b/pkg/yang/types_builtin.go
@@ -859,6 +859,7 @@ func parseRanges(s string, decimal bool, fracDigRequired uint8) (YangRange, erro
 		}
 		r[i] = YRange{min, max}
 	}
+	sort.Sort(r)
 	if err := r.Validate(); err != nil {
 		return nil, err
 	}
@@ -944,10 +945,9 @@ func (r YangRange) Less(i, j int) bool {
 	}
 }
 
-// Validate sorts r and returns an error if r has either an invalid range or has
+// Validate returns an error if r has either an invalid range or has
 // overlapping ranges.
 func (r YangRange) Validate() error {
-	sort.Sort(r)
 	switch {
 	case len(r) == 0:
 		return nil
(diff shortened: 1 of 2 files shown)


pkg/yang/types_builtin.go
```diff
@@ -859,6 +859,7 @@ func parseRanges(s string, decimal bool, fracDigRequired uint8) (YangRange, erro
 		}
 		r[i] = YRange{min, max}
 	}
+	sort.Sort(r)
 	if err := r.Validate(); err != nil {
 		return nil, err
 	}
```
wenovus, line 865: `coalesce` before `Validate`?

pkg/yang/types_builtin.go
```diff
@@ -944,10 +945,9 @@ func (r YangRange) Less(i, j int) bool {
...
 }
 
-// Validate sorts r and returns an error if r has either an invalid range or has
+// Validate returns an error if r has either an invalid range or has
 // overlapping ranges.
```
wenovus, line 949: Maybe just add at the end of the comment that it expects the range to be sorted.
Maybe also add to the docstring for `parseRanges`.

pkg/yang/types_builtin.go
```diff
@@ -944,10 +945,9 @@ func (r YangRange) Less(i, j int) bool {
...
 func (r YangRange) Validate() error {
-	sort.Sort(r)
```
wenovus, line 950: Hmm, now I'm thinking we should error out if https://golang.org/pkg/sort/#IsSorted returns false.
Also, could you add a `Sort` method for `YangRange`, and call that at `parseRanges`? Could you also add a short test with a couple test cases for the method?

This way if the user somehow manually creates a YangRange, he/she can call `.Sort` to allow it to be validated.

pkg/yang/types_builtin.go
harshilshah611: I forgot. Done

pkg/yang/types_builtin.go
harshilshah611: Added.

pkg/yang/types_builtin.go
harshilshah611: I see. Done

wenovus:
Thanks, looks good!

PR #122, opened 2020-05-15
Add Support for posix-pattern Extension within YangType

wenovus:
Added global variable `UsePosixPatternExt` to allow a `posix-pattern` extension from `openconfig-extensions.yang` to be used to replace pattern statements. This variable must be flipped prior to using goyang for consistent behaviour.

Restriction when `UsePosixPatternExt` is set:
- The number of `posix-pattern` statements must match the number of `pattern` statements within each type statement. Otherwise `goyang` parsing will fail.

### DIFF
--- a/pkg/yang/types.go
+++ b/pkg/yang/types.go
@@ -23,12 +23,6 @@ import (
 	"sync"
 )
 
-// UsePosixPatternExt specifies that openconfig-extensions:posix-pattern should
-// be used in place of the default pattern statement for all string types.  If
-// the number of posix-pattern statements doesn't match the number of pattern
-// statements within each type statement, the parsing will fail.
-var UsePosixPatternExt bool
-
 // A typeDictionary is a dictonary of all Typedefs defined in all Typedefers.
 // A map of Nodes is used rather than a map of Typedefers to simplify usage
 // when traversing up a Node tree.
@@ -352,13 +346,18 @@ check:
 	// Patterns are ANDed according to section 9.4.6.  If all the patterns
 	// declared by t were also declared by the type t is based on, then
 	// no patterns are added.
-	patterns := map[string]bool{}
+	seenPatterns := map[string]bool{}
 	for _, p := range y.Pattern {
-		patterns[p] = true
+		seenPatterns[p] = true
+	}
+	seenPOSIXPatterns := map[string]bool{}
+	for _, p := range y.POSIXPattern {
+		seenPOSIXPatterns[p] = true
 	}
 
-	checkAndAddPattern := func(n Node, p string) {
-		if _, err := syntax.Parse(p, syntax.Perl); err != nil {
+	checkPattern := func(n Node, p string, flags syntax.Flags) {
+		_, err := syntax.Parse(p, flags)
+		if err != nil {
 			if re, ok := err.(*syntax.Error); ok {
 				// Error adds "error parsing regexp" to
 				// the error, re.Code is the real error.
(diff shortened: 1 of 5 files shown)


pkg/yang/types.go
```diff
@@ -350,22 +356,40 @@ check:
 	for _, p := range y.Pattern {
 		patterns[p] = true
 	}
-	for _, pv := range t.Pattern {
-		p := pv.Name
+
+	checkAndAddPattern := func(n Node, p string) {
```
robshakir, line 360: I wonder whether this is the right place to do this parsing -- essentially, it means that downstream consumers never know about the extension, which I'm not sure is the right thing. Particularly, let's assume that there's some arbitrary tool that today does implement w3c patterns based on the output of goyang, in this model - if they specify `UsePosixPattern` because one module that they're using uses this form of pattern, then they will potentially get an error because there's no signal as to whether it was the POSIX pattern, or the regular pattern statement that was used to populate `(yang.Entry).Pattern`.

It seems like an alternate design would be to have a `POSIXPattern` field in the `Entry`, and then allow downstream tooling to determine which it uses.

pkg/yang/types.go
```diff
@@ -23,6 +23,12 @@ import (
 	"sync"
 )
 
+// UsePosixPatternExt specifies that openconfig-extensions:posix-pattern should
```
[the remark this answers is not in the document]
wenovus: Yes that's right... the only issue might be that the users' custom YANG files would need to be converted as well. However, I'm leaning towards not providing a transition exception file list because it just makes the tools more complex in having to support two different ways to parse files. I'd rather not have to deal with the correctness of that (e.g. augments, deviations, submodules, etc.)

pkg/yang/types.go
wenovus: Hmm I like that very much. It means the conversion has the steps of
1. update goyang, which doesn't break anything.
2. update models & downstream code.

I like it.

pkg/yang/types.go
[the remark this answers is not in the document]
wenovus: Yes that's right... I'm leaning towards not providing a transition exception file list because it just makes the tools more complex in having to support two different ways to parse files. Dealing with the correctness of that sounds complex (e.g. augments, deviations, submodules, etc.).

pkg/yang/types.go
```diff
@@ -351,23 +351,47 @@ check:
...
+			errs = append(errs, fmt.Errorf("%s: bad pattern: %v: %s", Source(n), err, p))
 		}
-		if !patterns[p] {
-			patterns[p] = true
-			y.Pattern = append(y.Pattern, p)
+	}
+
+	// First parse out the pattern statements.
+	for _, pv := range t.Pattern {
+		checkPattern(pv, pv.Name, syntax.Perl)
```
robshakir, line 377: The arguments to the `pattern` statement aren't actually `syntax.Perl` - but this looks to be as close as we could get. Maybe just add a comment here that says something like "we know that these are actually W3C patterns, but Golang doesn't support those, so we rather treat them like Perl patterns". That being said -- maybe we just don't check these patterns, since right now -- we'd likely error on cases that are specifically those that are in the W3C standard that golang's `syntax.Perl` doesn't recognise.

I like the idea of checking them for the `POSIX` patterns, I'm just worried about the fidelity problem above.

pkg/yang/types.go
```diff
@@ -351,23 +351,47 @@ check:
...
-	for _, pv := range t.Pattern {
-		p := pv.Name
-		if _, err := syntax.Parse(p, syntax.Perl); err != nil {
+	seenPOSIXPatterns := map[string]bool{}
+	for _, p := range y.POSIXPattern {
+		seenPOSIXPatterns[p] = true
+	}
+
+	checkPattern := func(n Node, p string, flags syntax.Flags) {
+		_, err := syntax.Parse(p, flags)
```
robshakir, line 364: `if _, err := syntax.Parse(p, flags); err != nil { ... }`

PR #123, opened 2020-06-03
Correctly inherit fraction-digit in typedef decimal64

wenovus:
Currently, whether a `decimal64` is a typedef is not detected if `t.FractionDigits` is unset in the typedef. So the ranges are being parsed as integer ranges, leading to https://github.com/openconfig/ygot/issues/397.

This change correctly identifies typedefs, and inherits the `FractionDigits` field correctly -- by not doing anything.

Note on the 3rd test case: whether overriding of `fraction-digits` is allowed is not specifically mentioned in the RFC. I followed `pyang`'s behaviour of erroring out if overriding is detected.

### DIFF
--- a/pkg/yang/entry_test.go
+++ b/pkg/yang/entry_test.go
@@ -2808,7 +2808,7 @@ func TestLeafEntryTypes(t *testing.T) {
 			}
 			`,
 		},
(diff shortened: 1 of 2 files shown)


wenovus:
Already covered in unit tests, integration test now added.

PR #130, opened 2020-06-28
Add support for multiple `base` substatements in `identity` statement

hslatman:
Squashed commit of the following:

commit 2cd89ac3d2344838867d446393e6e5174dca7700
Author: Herman Slatman <hermanslatman@hotmail.com>
Date:   Mon Jun 29 00:20:41 2020 +0200

    Fix existing tests that assume a single `base`

commit 528bac09ed575c111335a5d6b3589a5ab4e5c456
Author: Herman Slatman <hermanslatman@hotmail.com>
Date:   Mon Jun 29 00:14:23 2020 +0200

    Change `base` to be a slice

commit d1f5ccb6c756769739f92618e3a6149819a5526e
Author: Herman Slatman <hermanslatman@hotmail.com>
Date:   Mon Jun 29 00:08:06 2020 +0200

    Add support for multiple `base` substatements in `identity` statement

### DIFF
--- a/pkg/yang/identity_test.go
+++ b/pkg/yang/identity_test.go
@@ -33,10 +33,11 @@ type idrefOut struct {
 
 // identityOut is the output for a particular identity within the test case.
 type identityOut struct {
-	module   string   // The module that the identity is within.
-	name     string   // The name of the identity.
-	baseName string   // The base of the identity as a string.
-	values   []string // The string names of derived identities.
+	module    string   // The module that the identity is within.
+	name      string   // The name of the identity.
+	baseName  string   // The base of the identity as a string.
+	baseNames []string // The bases of the identity in test cases with multiple bases
+	values    []string // The string names of derived identities.
 }
 
 // identityTestCase is a test case for a module which contains identities.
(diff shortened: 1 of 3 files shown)


pkg/yang/identity_test.go
```diff
@@ -128,14 +128,14 @@ func TestIdentityExtract(t *testing.T) {
 			}
 
 			if ti.baseName != "" {
-				if ti.baseName != thisID.Base.Name {
+				if ti.baseName != thisID.Base[0].Name {
```
wenovus, line 131: Thanks for the contribution! Could you add a test case here for multiple base statements?

hslatman: I've added a test case for multiple bases.

In order to minimize changes to the existing test cases I added a new property `baseNames: []string` instead of changing the type of the existing `baseName` property in the test utility struct. If required, I can change that, of course.

wenovus: Thanks for adding the test case. Could you remove `baseName` please? It looks like just using `baseNames []string` can improve the code style.

I'm thinking of moving the `[]string` generation to a helper to make this easier, and then using the pattern here:
https://github.com/hslatman/goyang/blob/multiple-base-identities-1.1-v2/pkg/yang/node_test.go#L60
We're trying to use the new `cmp` package in tests whenever possible over `reflect.DeepEquals`.

hslatman: I've removed the `baseName` property in favor of `baseNames` and implemented the `cmp` package, resulting in a different kind of error message when the test fails, but I think it's more useful showing the actual diff. 

The utility function is usable in tests now. I don't know if this should be exposed in some way; as a method on `*Identity`, for example, or to just keep it like this.

wenovus: Nope, just keeping it here is fine, `cmp` is intended to only be used in tests.

wenovus:
Again this is a breaking change. @andaru @robshakir is it ok to merge it? LGTM.

PR #133, opened 2020-08-03
removed unnecessary type conversion

hellt:
It seems to me its safe to remove the type conversion in that case since the `data` is [already](https://github.com/openconfig/goyang/blob/b6e1e28273ad274f9240361b3a1ae65d27c045ac/pkg/yang/file.go#L91) a `string`.


andaru:
Thanks for posting this, I'd looked at (and over) this too many times!

PR #139, opened 2020-09-03
added support for leaf-list default substatement #136

### LINKED ISSUE #136: allow default substatments for leaf-list
Currently its not allowed, as it is not present in YANG 1.0 RFC,
YANG 1.1 adds `default` substatement to leaf-lists

hellt:
Hi @robshakir,
I've approached the issue described in #136 
Let me plz know if that PR satisfies the requirements of the feature and contribution guidelines.

closes #136 

-----------------
Cumulative summary after edits by @wenovus:
- `LeafList` now has a `Default []*Value` field.
- `Entry.Default` changed to []string .
- `Entry.DefaultValue() string` -> `Entry.SingleDefaultValue() (string, bool)` & `Entry.DefaultValues() []string`
  - Note that `Entry.SingleDefaultValue() (string, bool)` only returns true when there is one and only one default value. This includes the case when a leaf-list has a single default value.
- new field `YangType.HasDefault` to determine whether the empty string is a default value.

**NOTE**: This is a backwards-incompatible change (see summary)


hellt:
Hi @robshakir 
Did you have time to look into this PR? Anything I should address/fix?

pkg/yang/yang.go
```diff
@@ -421,6 +423,7 @@ type LeafList struct {
 	Extensions []*Statement `yang:"Ext"`
 
 	Config      *Value   `yang:"config"`
+	Default     *Value   `yang:"default"`
```
wenovus, line 426: Should this be an array since https://tools.ietf.org/html/rfc7950#section-7.7.2 says there could be multiple of these statements? `default` has cardinality 0..n. Admittedly `Deviate` also got this wrong, though is probably out of scope for this change.

wenovus (changes requested):

pkg/yang/yang.go
hellt: good find @wenovus, I will change the code accordingly

wenovus: I think if you just change `*Value` to `[]*Value` it might work. I think ast.go might be using `append()` to populate slice values.

hellt: @wenovus but how should I handle the `e.Default` assignment here - https://github.com/openconfig/goyang/pull/139/files#diff-c86a5ef7d87004179b80f56810efd550R568

entry's Default is a string; should I range over s.Default and create a `[]string` from Names of default statements which I later `strings.Join` to `e.Default`?

wenovus: Good point. I'm not sure which way to go for this issue. Your suggestion is straightforward; however, I don't see what delimiter to use to extract the individual values if we use `strings.Join` to create the value.

I see two options that may be more usable:
1. Add `Defaults []string` field for `Entry`, populated when it's a leaf-list.
 + Easier implementation.
 - Users of goyang will need to select the correct one.
2. Change the `Default` field to a slice type.
 + Single field for defaults. Users will simply use the `Entry.DefaultValue()` [method](https://github.com/openconfig/goyang/blob/633eccb6fa970a22464f8f9b22f23b26ff2487ed/pkg/yang/entry.go#L1441) to extract the singleton value.
 - Breaking change.

I'm kind of leaning towards 1. @robshakir @andaru would be happy to hear your thoughts.

wingeng: Could the `Defaults []string` be added to ListAttr?  From my reading of rfc7950, it's only in leaf-list that multiple defaults are allowed for a statement so might as well restrict this multi-default to that structure?

wenovus: Thanks @wingeng that sounds clever. However, that means the default value would be spread into two places.

@robshakir and I discussed, and decided to use option 2 coupled with the `SingleDefaultValue()` accessor. This way there is minimum chance of misuse for a new user.

hellt: @wenovus thanks for taking this offline. Do I understand correctly that someone from OC will change this PR to align it with the chosen method? Or should I take a shot at it?

wenovus: I will get to this eventually. However you're more than welcome to take a stab and we'd be willing to review.

hellt: Is there an example of that design approach that I can look at?

wenovus: Ok, I've implemented the option agreed on above.

One concern I have is that `SingleDefaultValue()` would return an empty string even if there was no default value. Since the default value *could* be the empty string, for example, we would not be supporting that.

As a result, I have added `HasDefault()` and `DefaultValues()` to help with checking for the existence of default values, and with getting all default values for a leaf-list type, respectively. `SingleDefaultValue()` would be useful for getting the default value for a non-leaf-list leaf entry.

wenovus:
leaf-list's default rules: https://datatracker.ietf.org/doc/html/rfc7950#section-7.7.2

pkg/yang/entry.go
```diff
@@ -1145,11 +1169,18 @@ func (e *Entry) ApplyDeviate() []error {
...
-						if devSpec.Default == deviatedNode.Default {
-							deviatedNode.Default = ""
-						} else {
-							appendErr(fmt.Errorf("%s: tried to deviate delete a default statement that doesn't exist or with a non-matching keyword", Source(e.Node)))
+					if len(devSpec.Default) > 0 {
+						switch {
+						case deviatedNode.IsLeafList():
+							// It is unclear from RFC7950 on how deviate delete works
+							// when there are duplicate leaf-list values.
+							appendErr(fmt.Errorf("%s: deviate delete on default statements unsupported for leaf-lists, please use replace instead", Source(e.Node)))
```
robshakir, line 1177: I'm not sure that I understand the error discussed above. Entries in a `leaf-list` must be unique ([ref](https://datatracker.ietf.org/doc/html/rfc7950#section-7.7)), and `delete` must match a specific keyword, so if you have:

```yang

leaf-list foo {
  type string;
  default "one";
  default "two";
}
```

then:

```yang

deviation "/foo" {
   deviate delete {
     default "one";
   }
}
```

then the default of `foo` would become `["two"]` rather than `["one", "two"]` in the before case. If there is no matching `default` for one that is removed then this is an error.

pkg/yang/entry.go
```diff
@@ -1519,17 +1550,40 @@ func errorSort(errors []error) []error {
...
-	} else if typ := e.Type; typ != nil {
-		if leaf, ok := e.Node.(*Leaf); ok {
-			if leaf.Mandatory == nil || leaf.Mandatory.Name == "false" {
-				return typ.Default
+// HasDefault indicates whether the leaf entry has a default value or not.
+func (e *Entry) HasDefault() bool {
+	return len(e.DefaultValues()) > 0
+}
+
+// SingleDefaultValue returns the schema default value for e, if any. This is
```
robshakir, line 1558: I wonder whether rather than requiring a user to do two calls here - i.e.,:

```golang

var default string
if a.HasDefault() {
   default = a.SingledefaultValue()
}
```

it's better to have these combined, such that you can do:

```
if def, ok := a.SingleDefaultValue(); a.IsLeaf() && ok {
   default = def
}
```

Since `DefaultValues` will return `nil`, then this is detectable in a single call for the `leaf-list` case so this approach would create some symmetry between the `leaf` and `leaf-list` types.

pkg/yang/entry_test.go
```diff
@@ -3173,6 +3338,29 @@ func TestDeviation(t *testing.T) {
...
+					}
+
+					deviation /a {
+						deviate delete {
+							default "fishsticks";
+						}
+					}
+				}`,
+		},
+		wantProcessErrSubstring: "deviate delete on default statements unsupported for leaf-lists",
```
robshakir, line 3363: Potentially we can remove this TODO and address the issue here if you agree with my assessment of 7950 above.

robshakir (changes requested):

pkg/yang/entry.go
wenovus: https://datatracker.ietf.org/doc/html/rfc7950#section-7.7
>    In configuration data, the values in a leaf-list MUST be unique.

What if it's state data? Then which one do we delete?

Unfortunately I couldn't find it in https://datatracker.ietf.org/doc/html/rfc7950#section-7.7.2, and when looking at pyang's behaviour, I hit this: https://github.com/mbj4668/pyang/issues/756.

pkg/yang/entry.go
wenovus: Doing it this way seems better for usability as well since it makes error handling more natural. I'll make the change.

pkg/yang/yang.go
wenovus: `HasDefault()` has been removed. See updated OP.

pkg/yang/entry.go
wenovus: Note that `Entry.SingleDefaultValue() (string, bool)` only returns true when there is one and only one default value. This includes the case when a leaf-list has a single default value.

pkg/yang/entry.go
wenovus: I agree that the behaviour can be implemented for config-false leaf-lists, so I've added a TODO to work on it after this PR. This PR is in someone else's repository so I prefer to open a PR for that after this is merged. We need to take care of erroring out when there are duplicate defaults.

pkg/yang/entry_test.go
wenovus: Can you see my reply above to see if you agree with what I intend to do for this here?

pkg/yang/entry.go
robshakir: That sounds fine with me.

This whole idea that `config true` and `config false` leaf-lists have different approaches seems a problem to me -- it just creates more complexity in the language. I'd be OK with an interpretation that just enforces uniqueness for config true and config false leaflists the same way.

pkg/yang/entry.go
robshakir: SGTM!

pkg/yang/entry_test.go
robshakir: Yep, that sounds good to me.

robshakir (approved):

wenovus (approved):
Thanks for the contribution @hellt!

PR #144, opened 2020-09-15
Add prefix support for RPC and Action input and output.

vingkaloom:
The Find method in entry.go does not support prefixes for input and output. Some ietf models such as [ietf-netconf-with-defaults@2011-06-01.yang](https://github.com/YangModels/yang/blob/cbfbdfcd445bb1fae2c029feea15441a9f9ce4eb/standard/ietf/RFC/ietf-netconf-with-defaults%402011-06-01.yang#L106) use a prefix for the input in an augment: `augment /nc:copy-config/nc:input`.


wenovus (approved):

wenovus:
Thanks for the change @vingkaloom!

PR #145, opened 2020-09-30
Detect missing closing brace from input

wingeng:
Consider the case of the following yang that is missing the final
closing brace for ```module foo```

```
module foo {
  namespace foo;
  prefix f;

  container top {
     leaf f1 {
        type string;
     }
  }

```

parsing that with ```./goyang -f tree test.yang``` results in no error
and no output.

With this patch we output an error that says we are missing a closing
brace.

### DIFF
--- a/pkg/yang/parse.go
+++ b/pkg/yang/parse.go
@@ -331,22 +331,20 @@ func (p *parser) nextStatement() *Statement {
 // Checks that we have a statement depth of 0. It's an error to exit
 // the parser with a depth of > 0, it means we are missing closing
 // braces. Note: the parser will error out for the case where we
-// start with an unmatched close brace
+// start with an unmatched close brace, eg. depth < 0
 //
 // This test is only done if there are no other errors as
 // we may exit early due to those errors -- and therefore there *might*
 // not really be a mismatched brace issue.
 func (p *parser) checkStatementDepth() {
-	// don't check if there are other errors
-	if p.errout.Len() > 0 {
+	if p.errout.Len() > 0 || p.statementDepth < 1 {
 		return
 	}
-	if p.statementDepth > 0 {
-		plural := ""
-		if p.statementDepth > 1 {
-			plural = "s"
-		}
-		fmt.Fprintf(p.errout, "%s:%d:%d: missing %d closing brace%s\n",
-			p.lex.file, p.lex.line, p.lex.col, p.statementDepth, plural)
+
+	plural := ""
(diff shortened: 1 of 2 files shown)


wingeng:
@googlebot I signed it!

pkg/yang/parse.go
```diff
@@ -321,3 +327,26 @@ func (p *parser) nextStatement() *Statement {
...
+	}
+	if p.statementDepth > 0 {
+		plural := ""
+		if p.statementDepth > 1 {
+			plural = "s"
+		}
+		fmt.Fprintf(p.errout, "%s:%d:%d: missing %d closing brace%s\n",
+			p.lex.file, p.lex.line, p.lex.col, p.statementDepth, plural)
+	}
+}
```
hellt, line 352: Hi @wingeng,
I am not a maintainer, but I'd write this piece with

* honouring early return in case there are no depth detected
* keeping happy path aligned on left

```go
func (p *parser) checkStatementDepth() {
	// don't check if there are other errors or no depth misalignment detected
	if p.errout.Len() > 0 || p.statementDepth == 0 {
		return
	}
	plural := ""
	if p.statementDepth > 1 {
		plural = "s"
	}
	fmt.Fprintf(p.errout, "%s:%d:%d: missing %d closing brace%s\n",
		p.lex.file, p.lex.line, p.lex.col, p.statementDepth, plural)
}
```

wingeng: @hellt, sure. Only one minor change to reflect the comment.

```
if p.errout.Len() > 0 || p.statementDepth < 1
```

Theres a case where there is a negative depth when we have too many closing braces .... although, now that I think about it, there should be an errout for that case.  Oh well, belts and suspenders.

wenovus: Since there is already a test for this I think it's fine.

pkg/yang/parse.go
```diff
@@ -35,6 +35,8 @@ type parser struct {
 	errout *bytes.Buffer
 	tokens []*token // stack of pushed tokens (for backing up)
 
+	statementDepth int
```
wenovus, line 38: Could you add a comment for this?

wenovus (approved):

pkg/yang/parse.go
wingeng: will do.

```
+       // Depth of statements in nested braces
        statementDepth int
```

wenovus:
Thanks!

PR #147, opened 2020-10-05
Add NewListAttr() function to construct ListAttr with MaxElements=math.MaxUint64

wenovus:
users of goyang can also make use of this call (e.g. ygot).

Having this function helps with test writing.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -142,7 +142,9 @@ type ListAttr struct {
 	OrderedBy   *Value // order of entries determined by "system" or "user"
 }
 
-func NewListAttr() *ListAttr {
+// NewDefaultListAttr returns a new ListAttr object with min/max elements being
+// set to 0/math.MaxUint64 respectively.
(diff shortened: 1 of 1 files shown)


pkg/yang/entry.go
```diff
@@ -142,6 +142,13 @@ type ListAttr struct {
 	OrderedBy   *Value // order of entries determined by "system" or "user"
 }
 
+func NewListAttr() *ListAttr {
```
robshakir, line 145: Can we add a documentation comment here please? It should likely explain that this gives a 0->max value so that users know that this is the case.

robshakir (changes requested):
Will we continue to have the same function in ygot as well as this goyang function? It looks like there'll be duplication for the moment.

pkg/yang/entry.go
wenovus: Done. Also changed name to `NewDefaultListAttr` as suggested elsewhere.

robshakir (approved):

Issue #148, opened 2020-10-08
sortedErrors Less method panics


karimra:
Hi,

I'm facing an issue with the method here: https://github.com/openconfig/goyang/blob/master/pkg/yang/entry.go#L1459

In the below code, the case where len(fi) == len(fi) == x, nless ends up panicking due to index out of range.
```
compare := func(x int) int {
	switch {
	case len(fi) == x && len(fj) > x:
		return -1
	case len(fj) == x && len(fi) > x:
		return 1
	case len(fj) < x && len(fi) < x:
		return 0
	}
	return nless(fi[x], fj[x])
}
```

In my use case, both `fi` and `fj` are identical with len == 1.

Also, I might be wrong but shouldn't `compare` take x (not 1) in the below code ?
```
for x := 1; x < 4; x++ {
	switch compare(1) {
	case -1:
		return true
	case 1:
		return false
	}
}
```

PR #149, opened 2020-10-14
Bugfix #148 sortedErrors less method panics

fredgan:


fredgan:
@wenovus @robshakir Anybody have a review?

pkg/yang/entry.go
```diff
@@ -1477,10 +1478,10 @@ func (s sortedErrors) Less(i, j int) bool {
 		case len(fj) < x && len(fi) < x:
```
wenovus, line 1478: This looks like a woefully under-documented part of the code, thanks for looking for a fix!

I think this line should be `case len(fj) == x && len(fi) == x:` to indicate that neither error slice has this substring index after the split. It's the same behaviour but more readable to me.

pkg/yang/entry.go
```diff
@@ -1477,10 +1478,10 @@ func (s sortedErrors) Less(i, j int) bool {
...
 			return 0
 		}
-		return nless(fi[x], fj[x])
+		return nless(fi[x-1], fj[x-1])
```
wenovus, line 1481: Can you add a comment here? Something like `// Both slices have this index: compare them.`

pkg/yang/entry.go
```diff
@@ -1477,10 +1478,10 @@ func (s sortedErrors) Less(i, j int) bool {
...
 	}
-	for x := 1; x < 4; x++ {
-		switch compare(1) {
+	for x := 1; x < errorSplitCount; x++ {
```
wenovus, line 1483: Can you add a comment here? Something like `// Compare remaining indices of the error string slices in order to create a total ordering.`

pkg/yang/entry.go
```diff
@@ -1457,8 +1457,9 @@ type sortedErrors []sError
 func (s sortedErrors) Len() int      { return len(s) }
 func (s sortedErrors) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
 func (s sortedErrors) Less(i, j int) bool {
-	fi := strings.SplitN(s[i].s, ":", 4)
-	fj := strings.SplitN(s[j].s, ":", 4)
+	const errorSplitCount = 4
```
wenovus, line 1460: Maybe add a comment here: `We expect the error strings to be composed of error messages, line numbers, etc. delimited by ":".`

pkg/yang/entry.go
```diff
@@ -1457,8 +1457,9 @@ type sortedErrors []sError
...
+	fi := strings.SplitN(s[i].s, ":", errorSplitCount)
+	fj := strings.SplitN(s[j].s, ":", errorSplitCount)
 	if fi[0] < fj[0] {
```
wenovus, line 1463: Can you add a comment here: `First, order the errors by the file name.`

pkg/yang/entry_test.go
```diff
@@ -3343,3 +3344,15 @@ func TestLeafEntryTypes(t *testing.T) {
 		})
 	}
 }
+
+func TestLess(t *testing.T) {
```
wenovus, line 3348: Thanks for adding a test! Can you make it more comprehensive: use table-based testing like the test above, and checking each edge case?

wenovus (changes requested):

pkg/yang/entry.go
fredgan: I don't think so. If this, it will crash because of out of index.

wenovus: Good point -- the algorithm will try again (since it returns 0) and error out next time. I thought it would magically exit the for loop when it sees this condition.

Maybe you can add a comment here `// If neither element exist, then they're equal`.

fredgan: Okey.  I've reread the code, in this case, I think it's better to return  `1` replaced ` 0` because it will exit the for loop immediately without going into the next loop.

wenovus: That makes sense. However, because doing this is somewhat unintuitive for the reader, because "1" here means >= instead of ==. So I would add to the comment above `compare := func(x int) int {` to say
`// -1 means less than, 1 means greater or *equal* to, and 0 means equal thus far, but still undecided and need to see the rest of the elements.

pkg/yang/entry.go
wenovus: nit: can you add a space after `//`?

pkg/yang/entry.go
wenovus: nit: can you add a space after `//`?

pkg/yang/entry_test.go
wenovus: Thanks for creating the cases, they look good. Could you put them into table format, like the test directly above?

This way you don't need to repeat the test logic.

wenovus:
Thanks again for the fix and increased coverage!

PR #150, opened 2020-10-21
The `mandatory` attribute on leafs was not handled correctly!

idefixcert:
The error was not covered in tests, because the deviation code had also an error.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -584,6 +584,8 @@ func ToEntry(n Node) (e *Entry) {
 		e.addError(err)
 		e.Prefix = getRootPrefix(e)
 		addExtraKeywordsToLeafEntry(n, e)
+		e.Mandatory, err = tristateValue(s.Mandatory)
+		e.addError(err)
 		return e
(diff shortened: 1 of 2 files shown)


wenovus (changes requested):
Thanks @idefixcert! Could you add a test for Entry? You can add a new test under `TestLeafEntryTypes`, and probably change the test name to something more general, like `TestLeafEntry`.

Could you also fix the current test for "deviation with delete" under `TestDeviation`? It looks to me like 
```
    container target {
        container delete {
            uses substmts;
        }
    }
```
has a test bug where reusing `substmts` is problematic as `leaf mandatory { type string; }` doesn't have `mandatory` set.

idefixcert:
Hey @wenovus, thanx for picking this up that fast.
I added two more test for mandatory and renamed the `TestLeafEntyTypes` test as you suggested.

But I did not get the point with the TestDeviation.
I am not so familiar with the deviation feature.
```grouping

wenovus (approved):
Ah I misread the test files as they're named so similarly. LGTM!

Issue #151, opened 2020-10-26
Unreasonable to print the union type


fredgan:
Hi,
It seems unreasonable to print the union type as `unionunion`, then I have a small change from `unionunion` to `union`, using yang module and test results as follows:

mod.yang:
```
module mod {
    namespace "urn:cent";
    prefix m;
        
    container bar {
        leaf my-mode {
            type union {
                type string;
                type enumeration {
                    enum short;
                    enum long;
                    enum wide;
                }
            }
        }
    }         
}
```
test results with `goyang -f types with mod.yang`:
```
#$goyang -f types mod.yang
unionunion{
    string;
    enumeration;
}    
```

fredgan:
Fixed by #152 Closing.

PR #152, opened 2020-10-26
Bugfix #151: Remove the redundant 'union' in the tree type output

fredgan:
Hi @wenovus Wenbo,
Currently goyang will print double "union" in the output with `-f type` option.
This PR remove the redundant "union" in the output.


wenovus (approved):
Looks good, thanks.

PR #153, opened 2020-11-11
The decimal's range should not be parsed if parsing fraction digits fails

fredgan:
…ails


pkg/yang/types.go
```diff
@@ -280,7 +280,8 @@ check:
 		y.IdentityBase = resolvedBase.Identity
 	}
 
-	if t.Range != nil {
+	// If parsing fractionDigits fails, the ranges should not be parsed
+	if t.Range != nil && isDecimal64 && y.FractionDigits != 0 {
```
robshakir, line 284: Could you check that the logic here is right? I think integer types may also have a range that isn't `nil`, but this resolution of the range is now only implemented for `decimal64`. If we could add a testcase to ensure that this is properly covered (and demonstrate that we didn't break this) that'd be great.

fredgan: You're right, parsing the range of integer types is ignored, I've adjusted the logic and added a testcase.

pkg/yang/types.go
```diff
@@ -280,7 +280,8 @@ check:
 		y.IdentityBase = resolvedBase.Identity
 	}
 
-	if t.Range != nil {
+	// If parsing fraction-digits of type decimal64 fails, the range should not be parsed
+	if t.Range != nil && (!isDecimal64 || (isDecimal64 && y.FractionDigits != 0)) {
```
wenovus, line 284: `parseRanges` already error checks where `FractionDigits` is 0 (via `decimalValueFromString`), so I'm wondering if it's actually better to keep the current behaviour and error out, instead of silently allowing parsing to continue?

fredgan: https://github.com/openconfig/goyang/blob/fd7d014c61f43a05d53146582e61503f374787df/pkg/yang/types.go#L247-L250
In fact, the value 0 indicates that the FractionDigits value has already failed to be parsed (via `asRangeInt`).

 If continue to parse, `parserRanges` checks the error value 0 as FractionDigits, then reports an unreasonable error msg `bad range: invalid number of fraction digits 0: out of range [1..18]`,  and the new testcase also fails because two errors are generated, so I think the value 0 needs to be excluded.

wenovus: Thanks! Makes sense.

wenovus:
I think this is reasonable to add, although I would want to wait until an integration test is added, which should be simple once #188 is merged.

PR #155, opened 2020-11-24
Error out on "max-elements 0"

### LINKED ISSUE #154: max-elements shouldn't allow 0
Currently, max-elements allows 0. This is not allowed by the RFC. I think we should do the following:
1. error out on 0.
2. Downstream code should treat 0 as `unbounded`. This doesn't make downstream code perfect, but is more robust, as 0 can be treated as `unbounded` instead of 0, which is fragile as it currently forces any instantiation of `ListAttr` to use `NewDefaultListAttr`.

wenovus:
Has to be positive per https://tools.ietf.org/html/rfc7950#section-7.7.6

Fixes #154

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -501,7 +501,10 @@ func semCheckMaxElements(v *Value) (uint64, error) {
 	}
 	val, err := strconv.ParseUint(v.Name, 10, 64)
 	if err != nil {
-		return val, fmt.Errorf(`%s: invalid max-elements value %q (expect "unbounded" or a non-negative integer): %v`, Source(v), v.Name, err)
+		return val, fmt.Errorf(`%s: invalid max-elements value %q (expect "unbounded" or a positive integer): %v`, Source(v), v.Name, err)
+	}
+	if val == 0 {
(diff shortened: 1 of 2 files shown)


pkg/yang/entry.go
```diff
@@ -501,7 +501,10 @@ func semCheckMaxElements(v *Value) (uint64, error) {
 	}
 	val, err := strconv.ParseUint(v.Name, 10, 64)
 	if err != nil {
-		return val, fmt.Errorf(`%s: invalid max-elements value %q (expect "unbounded" or a non-negative integer): %v`, Source(v), v.Name, err)
+		return val, fmt.Errorf(`%s: invalid max-elements value %q (expect "unbounded" or a positive integer): %v`, Source(v), v.Name, err)
+	}
+	if val == 0 {
+		return val, fmt.Errorf(`%s: invalid max-elements value %q (expect "unbounded" or a positive integer)`, Source(v), v.Name)
```
robshakir, line 507: Should we make this error message different to the one @504 -- it seems like them being the same might cause us some extra debugging effort in the future.

robshakir (approved):

PR #156, opened 2020-12-16
Add util package for goyang

wenovus:
The contents for `ProcessModules ` are copied from a private repository.


util/build_yang_test.go
```diff
@@ -0,0 +1,55 @@
+package util
```
robshakir, line 1: Should we move `util` to `pkg/util` -- since this is where we have the non-binary code in `goyang`?

robshakir, line 1: Can you add the copyright/license header in this file?

util/build_yang.go
```diff
@@ -0,0 +1,42 @@
+package util
```
robshakir, line 1: Same comment here re: adding a license header and copyright.

robshakir (approved):

PR #158, opened 2020-12-17
Add POSIXPattern to equal method

### LINKED ISSUE #157: goyang should not skip adding more string types to union if oc-ext:posix-pattern is not empty
If only posix-pattern is provided, then unions would be parsed incorrectly.

wenovus:
Fixes #157. Previously, if `pattern` is not present, `resolve()` in types.go thinks two string types are the same, even if `POSIXPattern` is different. This meant multiple leaves within a union containing only posix-pattern would be de-duped, and only the first pattern recorded.

Also improve `TestPattern` to be able to test the recursive structure of `Pattern` and `POSIXPattern` fields when the type is a union.

### DIFF
--- a/pkg/yang/types_test.go
+++ b/pkg/yang/types_test.go
@@ -283,6 +283,99 @@ func TestPattern(t *testing.T) {
 		wantPatternsRegular: []string{"alpha", "bravo", "charlie"},
 		wantPatternsPOSIX:   []string{"delta", "echo", "foxtrot"},
 	}, {
+		desc: "Union type -- de-duping string types",
+		leafNode: `
+			leaf test-leaf {
+				type union {
+					type string {
+						pattern 'alpha';
+						o:posix-pattern 'alpha';
+					}
+					type string {
+						pattern 'alpha';
+						o:posix-pattern 'alpha';
+					}
+				}
+			}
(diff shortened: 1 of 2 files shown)


pkg/yang/types_test.go
```diff
@@ -282,6 +282,99 @@ func TestPattern(t *testing.T) {
...
+				return nil, fmt.Errorf("node %v is missing imports", m)
+			}
+			e := ToEntry(m)
+			types := e.Dir["test-leaf"].Type.Type
+			if len(types) != 2 {
+				return nil, fmt.Errorf("Want 2 string entries, got %v types", len(types))
+			}
+			return types[1], nil
+		},
+		wantPatternsPOSIX: []string{"bravo"},
```
robshakir, line 377: A query about this test, should we just make `inGetFn` actually encode this testing, since here we're saying `[]strijng{"bravo"}` which is a bit counter intuitive since one would expect that we would check `["alpha", "bravo"]` as the patterns that are found.

wenovus: Thanks for the comment. I improved the test to be able to test the recursive structure of the pattern fields.

robshakir (approved):
LGTM, thanks for the change.

PR #160, opened 2021-02-08
Add Submodule identity test

wenovus:


pkg/yang/identity_test.go
```diff
@@ -131,6 +131,44 @@ var basicTestCases = []identityTestCase{
...
+					submodule idtest-one-sub {
+					  belongs-to idtest-one {
+					    prefix "idone";
+					  }
+
+					  identity TEST_ID;
+					}
+				`},
+		},
+		identities: []identityOut{
```
robshakir, line 163: To clarify the comment here -- the thing that doesn't show up is the `TEST_ID_DERIVED` identity?

In Y1.0, I believe the `include` statement that you have here should have made this valid, and in Y1.1, everything shares a namespace, so this should have been valid. From this I think that the Entry.merge should probably include the `Identities` field, and this was probably an oversight when I added support for the `identity` statement to goyang.

robshakir (changes requested):

pkg/yang/identity_test.go
wenovus: It's `TEST_ID` that doesn't show up. If you look at test case 3, you'll see that both the derived and non-derived identities show up if they're defined within the module file.

I've added another test case that shows when the base/derived identities in this test are reversed between the module and submodule, that now it's `TEST_ID_DERIVED` that no longer shows up.

The way I read the code is that goyang simplifies away submodules after parsing, so the merge should put the identities into the `Module` node. I'll open an issue.

wenovus: Upon checking, this isn't an issue in the `YangType` AST fields (which is more critical), and only the `Entry` AST node representing a module are affected. This is because the identity dict is still populated correctly by virtue of this: https://github.com/openconfig/goyang/blob/master/pkg/yang/identity.go#L154.

robshakir: Sorry to go back to an old discussion here -- given your comments above, I'd assume that both `TEST_ID` and `TEST_ID_DERIVED` are both meant to be listed here in all cases - they are both valid identities (albeit one doesn't have a base itself).

robshakir (changes requested):

pkg/yang/identity_test.go
wenovus: Yes I now agree this is still unresolved. I can make the change, but I would be changing the AST node (`Module`) object instead of the `Entry` object. Looking at the code for `Entry.merge`, it looks like the original intent is to keep the AST the same as the YANG textual representation, but make `Entry` the resolved representation. So, I'm wondering if it's better to keep the code as-is, but adding another test to test that the global (soon-to-be-local) identities dict actually contains all the resolved identities?

wenovus: If you agree then perhaps #104 should be merged first in order to do more easily test this currently-global identity dict.

robshakir: Yes, that makes sense to me (on both counts). Let me take a look at #104.

wenovus: Sorry I conflated #104's `typeDict` with the global `identities` dict. They actually have nothing to do with each other.. thanks for reviewing that change though, it definitely helped push the thinking on making goyang more than a one-time-use library.

I've now pushed a commit that tests using global identities dict instead of checking the parsed identities in the AST. The global `identities` dict is what is used when looking up any type that's an identityref`*`, so I believe this is a better test.

`*` https://github.com/openconfig/goyang/blob/master/pkg/yang/types.go#L144 `findIdentityBase` is the only function used to look-up identities, and `findIdentityBase` in turn looks them up from the global `identities` dict.

robshakir (approved):

wenovus:
Thanks for the reviews @robshakir!

PR #163, opened 2021-02-10
Clean-up Parser and Lexer

wenovus:
- Remove unused exported function `FakeStatement`.
- Simplify by replacing `closeBrace` and `openBrace` with their character values.
- Rename `tIdentifier` to `tUnquoted`. Identifiers can be quoted in YANG, so this name is confusing. The original author intended these to be unquoted strings.
- Improve comments in the parser.

### DIFF
--- a/pkg/yang/parse.go
+++ b/pkg/yang/parse.go
@@ -151,8 +151,8 @@ func (s *Statement) Write(w io.Writer, indent string) error {
 	return nil
 }
 
-// ignoreMe is an error recovery terminal used by the parser in
-// order to continue processing for other errors in the file.
+// ignoreMe is an error recovery non-terminal token used by the parser in order
+// to continue processing for other errors in the file.
 var ignoreMe = &Statement{}
 
 // Parse parses the input as generic YANG and returns the statements parsed.
--- a/pkg/yang/parse.go
+++ b/pkg/yang/parse.go
@@ -17,9 +17,6 @@ package yang
 // This file implements Parse, which  parses the input as generic YANG and
 // returns a slice of base Statements (which in turn may contain more
 // Statements, i.e., a slice of Statement trees.)
-//
-// TODO(borman): remove this TODO once ast.go is part of of this package.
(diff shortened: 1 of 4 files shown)


pkg/yang/parse.go
```diff
@@ -162,8 +151,8 @@ func (s *Statement) Write(w io.Writer, indent string) error {
 	return nil
 }
 
-// ignoreMe is returned to continue processing after an error (the parse will
-// fail, but we want to look for more errors).
+// ignoreMe is an error recovery non-terminal used by the parser in
```
robshakir, line 154: "error recovery non-terminal"? Should we say here "a non-terminal error that can be used in order...".

pkg/yang/parse.go
```diff
@@ -185,19 +174,18 @@ Loop:
...
 		case p.hitBrace:
-			fmt.Fprintf(p.errout, "%s:%d:%d: unexpected %c\n", ns.file, ns.line, ns.col, closeBrace)
+			fmt.Fprintf(p.errout, "%s:%d:%d: unexpected %c\n", ns.file, ns.line, ns.col, '}')
 		default:
 			statements = append(statements, ns)
 		}
 	}
 
-	p.checkStatementDepth()
+	p.checkStatementDepthIsZero()
```
robshakir, line 183: Do we need `check` in the name here? `statementDepthIsZero`?

robshakir (changes requested):

pkg/yang/parse.go
wenovus: Changed to "token" to describe this.

I think it's in the same spirit as the "error terminal" in the bison parser (last paragraph): https://www.gnu.org/software/bison/manual/html_node/Symbols.html

pkg/yang/parse.go
wenovus: I think the `check` is descriptive because it can output an error if the check fails.

robshakir (approved):

PR #164, opened 2021-02-10
Add tests for parser

wenovus:
`Parse` is now fully tested.


pkg/yang/parse.go
```diff
@@ -77,9 +77,8 @@ func (s *Statement) SubStatements() []*Statement { return s.statements }
 
 // String returns s's tree as a string.
```
robshakir, line 78: Best just to remove this given that it doesn't seem like it does anything particularly useful? Or is this part of the interface that needs to be implemented?

If it's not, should we create some simple implementation here (`YANG statement at line X`, or even. just `<YANG Statement>`)?

robshakir (approved):

pkg/yang/parse.go
wenovus: I'll delete this since it doesn't seem to be need by anything else in goyang as well as ygot.

If someone complains I will add an implementation for it. Added tracking bug at https://github.com/openconfig/goyang/issues/193

PR #170, opened 2021-02-17
Add identity.go tests

wenovus:
The if block checks for the base identity name using the non-prefix-qualified name, but such a name can't be found because the global `identities` dict is keyed by the prefix name (i.e. pfx:name), and not just the non-qualified name. There is no way the global dict contains the name, because `func (s *Identity) PrefixedName()`, which is used to create the key for the dict, will always print a ":" in the name, even if the prefix is somehow the empty string (disallowed by YANG but goyang doesn't seem to check for it), and `baseStr` does not contain the ':' character.

edit (21/07/08) -- the above statement is kept for historical reasons. It is not correct (see comments below).


pkg/yang/identity.go
```diff
@@ -95,13 +98,6 @@ func (mod *Module) findIdentityBase(baseStr string) (*resolvedIdentity, []error)
 			errs = append(errs, fmt.Errorf("%s: can't resolve the local base %s as %s", source, baseStr, keyName))
 		}
 	default:
-		// The identity we are looking for is prefix:basename.  If
```
robshakir, line 98: Just to clarify here -- is there any guarantee that this function gets an argument with `prefix:base`? 

I guess this statement can never be executed because if that were the case then `basePrefix` would be `""` such that this is unnecessary?

robshakir (approved):

pkg/yang/identity.go
```diff
@@ -25,7 +25,9 @@ import (
 // identityDictionary stores a global set of identities that have been resolved
 // to be identified by their module and name.
 type identityDictionary struct {
-	mu   sync.Mutex
+	mu sync.Mutex
+	// TODO(wenovus): The key should not be prefix:ident-name, but
+	// module-name:ident-name. Prefixes are NOT unique.
```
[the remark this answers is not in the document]
wenovus: Yep right here, filed against 1.0 https://github.com/openconfig/goyang/issues/173

pkg/yang/identity.go
wenovus: Thanks for looking at this, I must have read this code wrong the first time since this if block can only be called when the argument is `prefix:base`, and not anything else (notice the switch statement). This directly contradicts my PR description :facepalm: I'll revert back this snippet.

PR #181, opened 2021-06-19
Fix lexer to identify single-quote string that immediately occurs after a concatenation (plus sign).

### LINKED ISSUE #180: ietf-inet-types@2020-07-06.yang parsing
Hello,
thanks for this wonderful library.
we are using it for code generation.

I am struggling when parsing the ietf-inet-types@2020-07-06.yang file.

in particular:
```
  typedef email-address {
    type string {
      // dot-atom-text "@" ...
      pattern '[a-zA-Z0-9!#$%&'+"'"+'*+/=?^_`{|}~-]+'
            + '(\.[a-zA-Z0-9!#$%&'+"'"+'*+/=?^_`{|}~-]+)*'
            + '@'
            + '[a-zA-Z0-9!#$%&'+"'"+'*+/=?^_`{|}~-]+'
            + '(\.[a-zA-Z0-9!#$%&'+"'"+'*+/=?^_`{|}~-]+)*';
    }
    description
      "The email-address type represents an email address as
       defined as addr-spec in RFC 5322 section 3.4.1.";
    reference
      "RFC 5322: Internet Message Format";
  }
```

This causes 6 errors:
etf-inet-types@2020-07-06.yang:548:36: +'*+/=?^_`: syntax error
ietf-inet-types@2020-07-06.yang:548:46: {: not an identifier
ietf-inet-types@2020-07-06.yang:548:48: }: syntax error
ietf-inet-types@2020-07-06.yang:549:15: (\.[a-zA-Z0-9!#$%&': syntax error
ietf-inet-types@2020-07-06.yang:549:51: }: syntax error
ietf-inet-types@2020-07-06.yang:550:15: @[a-zA-Z0-9!#$%&': syntax error

But if I read https://datatracker.ietf.org/doc/html/rfc7950#section-6.1.3.1 correct than 
`'[a-zA-Z0-9!#$%&'+"'"+'*+/=?^_`{|}~-]+'`
are 3 strings concatenated by +.

wenovus:
This code makes the new assumption that a `'` that follows `+` must be a string.

This change will break YANG files where there exists an unquoted string that contains a `+'` character substring. Given that this is unlikely, I think this is a fine tradeoff to get cases like #180 working.

Indeed, the current lexer is already assuming `"` cannot exist within an unquoted string (see `lexIdentifier`). This is already a much more restrictive assumption.

NOTE: In YANG1.1:
```
An unquoted string cannot contain any single or double quote
      characters.
```

Fixes #180.


robshakir (approved):
Thanks for the fix Wen!

PR #182, opened 2021-06-19
Add more releases to goyang CI to keep in line with ygot

wenovus:


robshakir (approved):
Thanks for the catch!

PR #183, opened 2021-06-21
Add MatchingEntryExtensions() to search extensions added during uses/augment/include statements

wenovus:
`MatchingExtensions` currently takes in `yang.Node`, which is not quite usable when the user is mostly operating on `yang.Entry` objects.

I think the current `MatchingExtensions` implementation might still be useful in case the user wants to look into the extensions of non-Entry nodes, such as `yang.Type`.


robshakir (approved):
Thanks

PR #184, opened 2021-06-29
Modules.Parse errors out if ms.add(n) fails

### LINKED ISSUE #179: Modules.Parse shouldn't discard error from ms.add(n)
e.g. duplicate modules are detected, but simply ignored by the current implementation. This is actually a question in itself: should we error out or simply emit a warning?

wenovus:
If an error is detected during parsing (e.g. duplicate modules), the error should be returned to the user instead of ignored.

This is a backwards incompatible change.

Fixes #179.


pkg/yang/modules_test.go
```diff
@@ -29,6 +29,36 @@ var testdataFindModulesText = map[string]string{
 	"dup-ns-two":  `module dup-ns-two { prefix ns-two; namespace urn:duplicate; }`,
 }
 
+func TestDupModule(t *testing.T) {
+	tests := []struct {
+		desc      string
+		inModules map[string]string
+		wantErr   bool
+	}{{
+		desc: "two modules with the same name",
```
robshakir, line 38: If we are relying on `FindModuleByPrefix` at all, should we also make duplicate prefixes be reported as an error? In this case we would guarantee (for the moment) that prefixes are unique (although this is likely not robust as you point out).

robshakir (approved):

pkg/yang/modules_test.go
wenovus: I think I'd rather just go implement the correct resolution mechanism. The problem is not just that the prefixes the modules give themselves must be unique, but also that the importing modules must give a name that matches the suggested prefix (i.e. the prefix the modules give themselves).

Since doing this will only partially fix it, and I'm working on it this week and next, I think it's fine to skip this step.

PR #185, opened 2021-06-30
Minor style improvements

wenovus:
Add package comment and others.


pkg/util/build_yang.go
```diff
@@ -20,19 +22,19 @@ import (
 	"github.com/openconfig/goyang/pkg/yang"
 )
 
-// ProcessModules takes a list of modules, and a path specification and
-// runs the yang parser against them, returning a slice of yang.Entry
-// pointers which represent the top level modules that are to be parsed
-// by the struct generation.
-func ProcessModules(yangf, path []string) (map[string]*yang.Entry, []error) {
+// ProcessModules takes a list of either .yang file or module/submodule names
```
robshakir, line 25: Is this correct? I thought that the arguments here needed to be the paths to the `.yang` themselves for `ioutil.ReadFile` to be used.

Does this work if I just supply `openconfig-interfaces` for example, do we have test coverage for that?

wenovus: That's right, added a test for this case and re-wrote the docstring.

robshakir (approved):

PR #186, opened 2021-07-09
Change call location of initTypes for typeDictionary

wenovus:
Add `newTypeDictionary()` helper and call `initTypes` closer to where the dictionary is initialized.

Also change tests to use the new method to avoid using the global typeDict.


pkg/yang/bgp_test.go
```diff
@@ -28,7 +31,9 @@ func TestBGP(t *testing.T) {
 	if len(ss) != 1 {
 		t.Fatalf("got %d results, want 1", len(ss))
 	}
-	_, err = BuildAST(ss[0])
+	typeDict := newTypeDictionary()
+	initTypes(reflect.TypeOf(&meta{}), typeDict)
+	_, err = buildASTWithTypeDict(ss[0], typeDict)
```
robshakir, line 36:
```golang

if _, err := buildAST...; err != nil {
  ...
}
```

robshakir (approved):

pkg/yang/ast.go
```diff
@@ -78,6 +81,12 @@ type meta struct {
...
 // that parse (nearly) the same.
+// NOTE: This only works for root-level aliasing for now, which is good enough
+// for module/submodule. This is because yangStatement.funcs doesn't store the
+// handler function for aliased fields, and sRequired also may only store the
+// correct values when processing a root-level statement due to aliasing. These
+// issues would need to be fixed in order to support aliasing for non-top-level
+// statements.
 var aliases = map[string]string{
 	"submodule": "module",
 }
```
wenovus, line 92: @robshakir I had to change the test due to this change exposing a bug with the recently-merged aliasing test. Basically the test I added aliased a non-top-level node, but the code only works for a top-level one (module/submodule). I judged fixing the code to require some extra work so I instead changed the existing test to test aliasing on a top-level statement. I put this comment here in case someone comes across this again.

robshakir: SGTM. Thanks.

PR #187, opened 2021-07-15
Allow registering multiple enums within a union type.

### LINKED ISSUE #131: union []*YangType doesn't contain all enumeration subtypes
There is some incorrect behaviour where if there are multiple enumerations (typedef or not) within a union, then goyang fails to put all of them within the []*YangType field, meaning downstream tools will not recognize them.

There is some further strange behaviour where if these ignored enumerations are put within another level of nested union, some of them are then added to the AST. In any case, more testing and corresponding fixes are needed for this.

wenovus:
Previously `YangType.Equal` doesn't compare the `Enum` field, so it thought they were equal types and mistakenly de-duped them. This change simply makes that comparison happen to fix this issue.

Fixes #131 

### Impact on `ygot`
[This TODO](https://github.com/openconfig/ygot/blob/master/testdata/modules/enum-types.yang#L51-L52) is now **partially** fixed. The second enumeration is still not recognized by `ygot`, but `td2` and `td-enum` are now both recognized. Ignoring the second enumeration may actually be a good option, since it avoids the issue of how to handle name and value conflicts between two enum values, which per the [RFC](https://datatracker.ietf.org/doc/html/rfc7950#section-9.6.4) both have to be unique. Ignoring any second enumerations also doesn't at all reduce the expressiveness of YANG. `pyang` is even blind to these conflicts since it doesn't flag it.

Where the second enumeration is being ignored (L531): https://github.com/openconfig/ygot/blob/d9bb85086db130b5e12fa93779dbc72e1a0dea5d/ygen/enumgen.go#L521-L534


pkg/yang/yangtype.go
```diff
@@ -266,11 +270,14 @@ func (y *YangType) Equal(t *YangType) bool {
 		!ssEqual(y.POSIXPattern, t.POSIXPattern),
 		len(y.Range) != len(t.Range),
 		!y.Range.Equal(t.Range),
-		!tsEqual(y.Type, t.Type):
+		!tsEqual(y.Type, t.Type),
+		!cmp.Equal(y.Enum, t.Enum, cmp.Comparer(func(t, u EnumType) bool {
```
robshakir, line 274: This is interesting -- prior to this change, I presume that we were de-duplicating enumerations because we weren't comparing the actual `EnumType` that was within the `YangType`, is that right?

robshakir (approved):

pkg/yang/yangtype.go
wenovus: Yes it doesn't check either the name nor the definition of the enum, so the comparison might return true and it gets discarded.

PR #188, opened 2021-07-19
Inherit min/max in already-restricted range/length statements.

### LINKED ISSUE #128: range statement verification handles `min` and `max` incorrectly
`goyang` currently has a bug whereby it always recognizes `min` and `max` as the min and max of the root type (i.e. uint64, int8, etc.) instead of its [direct typedef parent](https://tools.ietf.org/html/rfc7950#section-9.2.5). This leads to an incorrect (overly-lenient) verification of YANG types during AST construction ([example file](https://github.com/openconfig/goyang/files/4828305/test.yang.txt)), and to an incorrect (overly-lenient) validation of actual values for downstream libraries such as `ygot` when `min` or `max` are not -inf and +inf.

To fix this, each `Number` simply needs to know the actual values of `min` and `max`, and the easiest way to do that would be to abolish `MinNumber`/`MaxNumber` ([link](https://github.com/openconfig/goyang/blob/master/pkg/yang/types_builtin.go#L503)), and use the actual values instead.

This is a backwards-incompatible change, but just like #125, I believe the usability improvement is worth the cost, as it also fixes bugs.

The proposed changes have the following backwards-incompatible components:
- `MaxNumber`/`MinNumber` will be removed as `Number` types. All numbers will be concrete values once parsed into a `Number` type.

- `ParseInt`, `ParseDecimal`, `ParseRangesInt`, and `ParseRangesDecimal` will now take min/max as function parameters to resolve `min` and `max` inputs to concrete values when parsing.

It fixes the following:
- `YangRange.Contains` will filter out invalid YANG that currently would pass.
- `YangRange` will store only concrete values, making range validations much easier for downstream libraries such as `ygot`.

wenovus:
- Definition of `Number` is changed (**backwards-incompatible**).
- `MaxNumber`/`MinNumber` are removed as Number types. All numbers are concrete values once parsed into a Number type.
- When parsing range/length statements, a parent range is required in
order to infer the values of min/max (`parseChildRanges()`).

**NOTE**: This is a backwards-incompatible change due to the removal of
`Kind` from the definition of `Number`.

Testing:
- New unit tests.
- Integration tests added for inheritance.

Fixes #128


pkg/yang/types_builtin.go
```diff
@@ -40,14 +40,17 @@ var (
 	Uint32Range = mustParseRangesInt("0..4294967295")
 	Uint64Range = mustParseRangesInt("0..18446744073709551615")
 
-	Decimal64Range = mustParseRangesDecimal("min..max", 1)
+	Decimal64Range = mustParseRangesDecimal("-922337203685477580.8..922337203685477580.7", 1)
```
robshakir, line 43: Given that the range for `decimal64` differs based on the precision, is there something that we need to do to actually check the range within the precision that is specified? I suspect the answer here is yes, and that we rather need to make the maximum values dependent on the fraction digits -- one of the unit tests expects that this range can be used with `fraction-digits 2` which isn't allowed.

wenovus: Thanks for catching this! I almost introduced a new bug. I missed that `fraction-digit` is a required statement, so if we can't use min/max for the base type, then we must populate the `Range` field once we know its value. I've done that now, PTAL.

pkg/yang/types_builtin.go
```diff
@@ -361,24 +300,32 @@ func (r YangRange) Contains(s YangRange) bool {
...
-// larger than MaxInt64/MinInt64 are converted into maxNumber/minNumber.
+// outside the boundaries specified for the decimal64 value specified in
+// RFC6020/RFC7950 are clamped down to the closest boundary value.
 func FromFloat(f float64) Number {
-	if f > float64(MaxInt64) {
-		return maxNumber
+	if f > MaxDecimal64 {
+		return Number{
+			Value:          FromInt(MaxInt64).Value,
+			FractionDigits: 1,
```
robshakir, line 320: i'm not sure I understand here. if `f > MaxDecimal64` shouldn't it be impossible to create this value - since we don't have any floating-point type in yang, and we're creating this `Number` in the context of a yang schema, what is the scenario in which it is not an error to create a greater-than-maximum-decimal64 `Number` that we're handling here?

pkg/yang/types_builtin.go
```diff
@@ -361,24 +300,32 @@ func (r YangRange) Contains(s YangRange) bool {
...
 // FromFloat creates a Number from a float64. Input values with absolute value
-// larger than MaxInt64/MinInt64 are converted into maxNumber/minNumber.
+// outside the boundaries specified for the decimal64 value specified in
+// RFC6020/RFC7950 are clamped down to the closest boundary value.
 func FromFloat(f float64) Number {
-	if f > float64(MaxInt64) {
-		return maxNumber
+	if f > MaxDecimal64 {
+		return Number{
+			Value:          FromInt(MaxInt64).Value,
```
robshakir, line 319: can you clarify why this is `MaxInt64` rather than whatever was in `f`? It seems like we return a number here that is being treated as magic?

pkg/yang/types_builtin.go
wenovus: This function is only used by tests in goyang and ygot.

The primary reason why I chose to not return an error is because it makes it easier on tests that use this function, since you can just declare it within a struct declaration. Since ygot uses this, I can't make this function unexported without having to copy it into a different package, which I think is not ideal either.

robshakir (approved):

PR #191, opened 2021-07-28
Fix deviation statements on default statements.

wenovus:
Currently, `deviate add` and `deviate delete` on defaults statements are implemented incorrectly, this change makes the necessary fixes.


robshakir (approved):
LGTM, thanks for the change.

PR #192, opened 2021-07-28
Fix merge conflict between #163 and #181

wenovus:
@robshakir, I didn't do a merge on the older PRs before merging, leading to this breakage on HEAD. Is there a setting in goyang you can enable to make it similar ygot where you need to merge into master first before merge?


robshakir (approved):
LGTM, and I added the same review and requirements in this repo as ygot.

PR #199, opened 2021-08-20
Simplify some code relating to prefixes.

wenovus:


pkg/yang/find.go
```diff
@@ -21,33 +21,28 @@ import (
...
-	for {
-		if m, ok := n.(*Module); ok {
-			if m.Prefix != nil && m.Prefix.Name == parts[0] {
-				return parts[1]
-			}
-			return name
-		}
-		n = n.ParentNode()
+func trimLocalPrefix(n Node, name string) string {
+	pfx := localPrefix(n)
```
robshakir, line 33: Can `localPrefix` ever return nil? It seems like it'd be an error to have no prefix defined? (I'm OK that we don't handle it, I'm just trying to work out when the below condition is required.

robshakir: Oh, I see - you're expecting that `name` includes something like "foo:bar", and this is to trim `foo:`. 

Could you update the docstring to say that if the name is not prefixed with the local module, or is unprefixed entirely, then the same thing is returned?

pkg/yang/find.go
```diff
@@ -63,8 +58,6 @@ func FindGrouping(n Node, name string, seen map[string]bool) *Grouping {
 		v = e.FieldByName("Import")
 		if v.IsValid() {
 			for _, i := range v.Interface().([]*Import) {
-				// TODO(borman): This is arguably wrong.  We
-				// need to do prefix matching.
 				pname := strings.TrimPrefix(name, i.Prefix.Name+":")
```
robshakir, line 61: So just to be clear, we're now saying this is not wrong - because when we do any `FindXXX` then we'll support using the prefix within the context of the module that we found it?

robshakir (approved):
A couple of questions just to ensure that I understood this fully, but this LGTM.

pkg/yang/find.go
wenovus: I think this wasn't wrong in the first place because `name` MUST be either prefixed with the same prefix from a particular import statement, or otherwise it's not from the module. Thus we trim it before searching. Otherwise we skip it (`if pname == name`).

wenovus: I'll add a comment, I had to re-read the code to figure this out again.

PR #200, opened 2021-08-20
Depreciate Modules.FindModuleByPrefix method.

wenovus:
This method doesn't make sense because prefixes may be duplicated.

Instead, searches based on the prefix alone should be based on the textual location of an AST node, whether it is an extension, uses, augment, deviation, type, or leafref, so use the `FindModuleByPrefix` function instead.


robshakir (approved):
Great, thanks for this Wen. This looks a tonne tidier!

r.

PR #201, opened 2021-08-20
Simplify resolved identity look-up by using mname:basename and support YANG1.1 scoping.

### LINKED ISSUE #173: Implement Module-Dependent Prefix Resolution
As mentioned in comment in the codebase already, goyang is assuming in some places that prefixes are unique across all modules and will stay the same in all importing modules compared to the original module. This is incorrect, and module-dependent prefix resolution should be implemented.

wenovus:
Fixes #173.

It was previously caching `prefix:basename`, and if there is a non-matching name, it would then do the correct prefix-resolution and then another loop across the identities.

Since `modulename:basename` is being used, I've removed this now redundant second loop.

Furthermore, the old resolution mechanism didn't allow searching for a name defined outside of the module/submodule file, which the current resolution mechanism now supports (`test-case-0`).

### DIFF
--- a/pkg/yang/identity.go
+++ b/pkg/yang/identity.go
@@ -29,7 +29,7 @@ type identityDictionary struct {
 	// dict is a global cache of identities keyed by
 	// modulename:identityname, where modulename is the full name of the
 	// module to which the identity belongs. If the identity were defined
-	// in a submodule, then the belonging module name is used instead.
+	// in a submodule, then the parent module name is used instead.
 	dict map[string]resolvedIdentity
 }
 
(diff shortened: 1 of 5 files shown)


pkg/yang/identity.go
```diff
@@ -26,8 +26,10 @@ import (
 // to be identified by their module and name.
 type identityDictionary struct {
 	mu sync.Mutex
-	// TODO(wenovus): The key should not be prefix:ident-name, but
-	// module-name:ident-name. Prefixes are NOT unique.
+	// dict is a global cache of identities keyed by
+	// modulename:identityname, where modulename is the full name of the
+	// module to which the identity belongs. If the identity were defined
+	// in a submodule, then the belonging module name is used instead.
```
robshakir, line 32: (nit) `parent module name is used instead`.

robshakir (approved):
Again, this is a lot cleaner. I'd forgotten that we'd added specific handling here.

Thanks for the change.

pkg/yang/identity.go
wenovus: Done.

PR #202, opened 2021-08-24
Ignore extra error tokens.

### LINKED ISSUE #116: error channel is not consumed

wenovus:
An alternative that fixes #116.

This is essentially what @pborman proposed in https://github.com/openconfig/goyang/pull/117#issuecomment-616119988.

I wasn't able to reproduce test failing as mentioned by @andaru in https://github.com/openconfig/goyang/pull/117#issuecomment-622232317 in the current codebase, so pborman's suggestion now makes more sense.

The reason seems to be that error tokens are always skipped, so they don't seem to affect `Location` as Andrew mentioned:
https://github.com/openconfig/goyang/blob/1fb7893798e04830b4d7fbbcd6672b95668fcf73/pkg/yang/parse.go#L207-L214

### DIFF
--- a/pkg/yang/lex.go
+++ b/pkg/yang/lex.go
@@ -145,7 +145,7 @@ func newLexer(input, path string) *lexer {
 		file:   path,
 		input:  input,
 		line:   1, // humans start with 1
-		items:  make(chan *token, 3),
+		items:  make(chan *token, maxErrors),
 		state:  lexGround,
 		errout: os.Stderr,
 	}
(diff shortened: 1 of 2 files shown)


robshakir (approved):
Thanks - I'm not sure that we ever properly reproduced this issue -- but this looks to handle the invalid escapes that were being discussed.

Issue #172, opened 2021-02-18
goyang's IR generation should be hermetic


wenovus:
Currently, goyang does not safely allow multiple sets of modules to be processed within a single process instance. The user has to fire up a different process in order to parse a different set of files, especially if the files use a modified version of a module.

The task is to put any global options and cache variables in a struct, and for that struct's methods to do processing of files. This work would most likely modify the `Modules` struct modules.go to achieve this to minimize backwards-incompatible changes.

wenovus:
### Resolved global variables/options
- `typeDict`: https://github.com/openconfig/goyang/pull/104

### Unresolved
file.go
- `Path` and `pathMap`

identities.go
- `identities`

options.go
- `ParseOptions`

### Unresolved but not impactful
entry.go
- `entryCache` and `mergedSubmodule`. These are cleared at every `Modules.Process`.

The ones below this point are dependent on either an `init` function, or what's passed to `initTypes`, though `initTypes` is always called by `Modules` with the same argument so it's always initialized to the same values.

ast.go
- `typeMap`, `nameMap`, `aliases`. 

types_builtin.go
- `BaseTypedefs`

wenovus:
Action plan for "unresolved but not impactful" items:
- `BaseTypedefs` doesn't ever need to be changed since it's only called by `init()` and therefore always going to be the same.
- entry.go's cached variables should be changed to not create thread-safety issues.
- The other  (`typeMap`, `nameMap`, `aliases`) can remain global variables; however, we should factor out `d *typeDictionary` as a parameter to `initTypes`, and also call `initTypes` on `init()` only. This would make these always the same, and therefore shareable between different `Modules` objects.

wenovus:
- `identities` can be stored in `typeDictionary` (backwards compatible in practice).
- `entryCache` and `mergedSubmodule` are easy to fix -- just put them in `Modules` (backwards compatible).
- `ParseOptions` can just go in `Modules` (backwards incompatible).

PR #204, opened 2021-08-30
factor out `d *typeDictionary` as a parameter to `initTypes`.

wenovus:
Also call `initTypes` on `init()` only. This confirms that the global
variables `typeMap`, `nameMap`, `aliases` are always the same values, so
that they don't need to be created per instance of the `Modules` object.

This refactor is preparation work for #172 (allowing multiple sets of modules to be processed independently).

### DIFF
--- a/pkg/yang/ast.go
+++ b/pkg/yang/ast.go
@@ -31,6 +31,10 @@ import (
 )
 
 func init() {
+	// Initialize the global variables `typeMap` and `nameMap`.
+	// By doing this, we are making the assumption that all modules will be
+	// parsed according to the type hierarchy rooted at `meta`, and thus
+	// all input YANG modules will be parsed in this manner.
 	initTypes(reflect.TypeOf(&meta{}))
 }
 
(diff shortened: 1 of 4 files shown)


pkg/yang/ast.go
```diff
@@ -30,12 +30,16 @@ import (
 	"strings"
 )
 
+func init() {
+	initTypes(reflect.TypeOf(&meta{}))
```
robshakir, line 34: Add a comment here just to mention what we're doing here - since I guess you're going to move this out of `init` when we have a context within which modules are generated?

pkg/yang/ast.go
```diff
@@ -143,13 +147,13 @@ func build(s *Statement, p reflect.Value, d *typeDictionary) (v reflect.Value, e
 
 	if fn := y.funcs["Name"]; fn != nil {
 		// Name uses s directly.
-		if err := fn(s, v, p); err != nil {
+		if err := fn(s, v, p, d); err != nil {
```
robshakir, line 150: I think the names here have got a bit too pithy -- can we call them something a bit more descriptive, I'd suggest:

```
s --> stmt
p --> parent
d --> types
```

`d` particularly doesn't make much sense when you're reading through.

robshakir (changes requested):

pkg/yang/ast.go
wenovus: Added comment, although I'm not planning on moving it out of `init`. Please see OP.

pkg/yang/ast.go
wenovus: Done.

pkg/yang/ast.go
robshakir: maybe I'm missing something on the design here.

when we have a context for the generation of modules, I presume all the context for any generation will be within the `Modules` parent (or whatever you call `Parse()` on). is that incorrect? I'm struggling a little to see why we need to initialise a global in `init()` if we're not planning to have globals, but I expect I'm missing something here.

wenovus: Under the proposal by this PR we will have globals. It's just that the globals will be constant. i.e. variables that won't ever change, namely the three global variables mentioned in the OP.

The alternative would be to store these variables in `Modules` as well, but I don't see the need because as long as we don't have another type hierarchy (currently rooted at `meta`), for example if we wanted to be able to parse YANG1.1 separately, then there is no need to initialize these global variables again and again each time we instantiate a `Modules` object. Let me know if this made sense.

robshakir: SGTM, makes entire sense.

robshakir (approved):

PR #208, opened 2021-08-30
Move global ParseOptions into Modules.

wenovus:
Part of the effort of #172.
**Backwards-incompatible.**

### DIFF
--- a/pkg/yang/modules.go
+++ b/pkg/yang/modules.go
@@ -31,17 +31,27 @@ type Modules struct {
 	byPrefix   map[string]*Module // Cache of prefix lookup
 	byNS       map[string]*Module // Cache of namespace lookup
 	typeDict   *typeDictionary    // Cache for type definitions.
+	// entryCache is used to prevent unnecessary recursion into previously
+	// converted nodes.
+	entryCache map[Node]*Entry
+	// mergedSubmodule is used to prevent re-parsing a submodule that has already
+	// been merged into a particular entity when circular dependencies are being
+	// ignored. The keys of the map are a string that is formed by concatenating
+	// the name of the including (sub)module and the included submodule.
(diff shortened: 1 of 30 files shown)


robshakir (changes requested):
I think this is going to be a breaking change AFAICS -- since anyone setting the `ParseOptions` directly will now fail to build.

Can we tag a version before this one is merged so that folks can pin to that version before they take this change on please?

robshakir:
Oh, sorry - I missed the backwards-incompatible in the description. Same comment above still applies.

robshakir:
@wenovus - Can you check whether you're waiting on something from me here? thanks!

pkg/yang/modules.go
```diff
@@ -39,7 +39,12 @@ type Modules struct {
 	// ignored. The keys of the map are a string that is formed by concatenating
 	// the name of the including (sub)module and the included submodule.
 	mergedSubmodule map[string]bool
-	mu              sync.Mutex // Mutex to protect byNS map
+	// ParseOptions sets the options for the current YANG module parsing. It can be
+	// directly set by the caller to influence how goyang will behave in the presence
+	// of certain exceptional cases.
+	ParseOptions Options
+	// Mutex to protect byNS map
+	mu sync.Mutex
```
robshakir, line 47: It's idiomatic to have the mutex defined above the value that is being protected - so here:

```golang
nsMu sync.Mutex // ...
byNS map[string]*Module // ...
```

I suggest calling it `nsMu` since I'd expect `mu` would protect all of the `map` types in the struct.

robshakir (approved):

pkg/yang/modules.go
wenovus: Done.

PR #212, opened 2021-09-12
Adding Extras to JSON - fixes #211

### LINKED ISSUE #211: Extras are excluded from JSON output
The line
```go
Extra map[string][]interface{} `json:"-"`
```
means that all Extras are excluded from JSON output.

This was probably done because the JSON marshaler gives an error like:
```
ERROR Generating GoStruct Code: error marshalling JSON schema: JSON marshalling error: json: unsupported value: encountered a cycle via *yang.Module
```

This can be easily avoided by adding `json:"-"` to the `Parent` field of any struct that is used as an Extra

SeanCondon:
In this Patch I'm doing 2 things
1) Removing the exclusion of Extras from JSON, and excluding down the line the Parent and Source of Extra elements

2) Improving how items are added to the Extras slice. When the item is a Slice add its elements individually.

I can break this up in to 2 separate patches if necessary

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -120,7 +120,7 @@ type Entry struct {
 	Uses            []*UsesStmt `json:",omitempty"` // Uses merged into this entry.
 
 	// Extra maps all the unsupported fields to their values
-	Extra map[string][]interface{} `json:",omitempty"`
+	Extra map[string][]interface{} `json:"extra-unsupported-unstable,omitempty"`
 
 	// Annotation stores annotated values, and is not populated by this
 	// library but rather can be used by calling code where additional
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -120,7 +120,7 @@ type Entry struct {
 	Uses            []*UsesStmt `json:",omitempty"` // Uses merged into this entry.
 
 	// Extra maps all the unsupported fields to their values
-	Extra map[string][]interface{} `json:"-"`
+	Extra map[string][]interface{} `json:",omitempty"`
 
 	// Annotation stores annotated values, and is not populated by this
 	// library but rather can be used by calling code where additional
@@ -980,7 +980,9 @@ func ToEntry(n Node) (e *Entry) {
 			"unique",
 			"when",
 			"yang-version":
-			e.Extra[name] = append(e.Extra[name], fv.Interface())
+			if !fv.IsNil() {
+				addToExtrasSlice(fv, name, e)
+			}
 			continue
 
 		case "Ext", "Name", "Parent", "Statement":
@@ -1024,8 +1026,27 @@ func addExtraKeywordsToLeafEntry(n Node, e *Entry) {
 			"reference",
 			"status",
 			"when":
-			e.Extra[name] = append(e.Extra[name], fv.Interface())
+			if !fv.IsNil() {
+				addToExtrasSlice(fv, name, e)
+			}
 		}
+	}
+}
+
+func addToExtrasSlice(fv reflect.Value, name string, e *Entry) {
+	if fv.Kind() == reflect.Slice {
+	outerLoop:
+		for j := 0; j < fv.Len(); j++ {
+			for _, en := range e.Extra[name] {
+				if en == fv.Index(j).Interface() {
+					// Don't add again if identical
+					continue outerLoop
+				}
+			}
+			e.Extra[name] = append(e.Extra[name], fv.Index(j).Interface())
+		}
+	} else {
+		e.Extra[name] = append(e.Extra[name], fv.Interface())
 	}
 }
 
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -120,7 +120,7 @@ type Entry struct {
 	Uses            []*UsesStmt `json:",omitempty"` // Uses merged into this entry.
 
 	// Extra maps all the unsupported fields to their values
-	Extra map[string][]interface{} `json:"extra-unsupported-unstable,omitempty"`
+	Extra map[string][]interface{} `json:"extra-unstable,omitempty"`
 
 	// Annotation stores annotated values, and is not populated by this
 	// library but rather can be used by calling code where additional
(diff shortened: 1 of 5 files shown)


SeanCondon:
Hi @wenovus - I hope you will get a chance to look at this in the near future - best regards, Sean

robshakir (commented):
One concern here - should we be making `Extras` be something that is a public API that can be relied upon? It seems like the unimplemented keywords that we have specified should probably just be implemented as a use-case comes up for them.

Is there a specific case that this is being implemented for?

SeanCondon:
> One concern here - should we be making `Extras` be something that is a public API that can be relied upon? It seems like the unimplemented keywords that we have specified should probably just be implemented as a use-case comes up for them.
> 
> Is there a specific case that this is being implemented for?

It was really the `must` statement I was interested in.
So yes there could be a direct object containing the must statments.

As a YANG author though I'd like to see the statement I've written to be in the JSON somehow. This is why I thought exposing the Extras has value.

wenovus:
I agree that if we introduced this change, that dependent users would see an incompatible change if we later add support for some of the statements.

As a compromise we could marshal the field into a JSON field with a name that is unambiguously unstable.

```go
Extra map[string][]interface{} `json:"unsupported-unstable,omitempty"` 
```

SeanCondon:
> ```go
> unsupported-unstable
> ```
Thanks @wenovus and @robshakir for reviewing.
Re the `unsupported-unstable` tag - could I suggest `extra-unsupported-unstable`, so that it's clear it represents the `extra` section.

pkg/yang/entry.go
```diff
@@ -1024,11 +1026,30 @@ func addExtraKeywordsToLeafEntry(n Node, e *Entry) {
...
 			"when":
-			e.Extra[name] = append(e.Extra[name], fv.Interface())
+			if !fv.IsNil() {
+				addToExtrasSlice(fv, name, e)
+			}
 		}
 	}
 }
 
+func addToExtrasSlice(fv reflect.Value, name string, e *Entry) {
```
wenovus, line 1036: I wonder why this processing is necessary?

pkg/yang/entry.go
```diff
@@ -1024,11 +1026,30 @@ func addExtraKeywordsToLeafEntry(n Node, e *Entry) {
...
+	if fv.Kind() == reflect.Slice {
+	outerLoop:
+		for j := 0; j < fv.Len(); j++ {
+			for _, en := range e.Extra[name] {
+				if en == fv.Index(j).Interface() {
+					// Don't add again if identical
+					continue outerLoop
+				}
```
wenovus, line 1044: What's the rationale for removing duplicates? Maybe a particular extension has a different meaning when duplicated?

pkg/yang/entry.go
wenovus: Mainly this is non-backwards-compatible. I'm wondering if this can be done a layer up?

SeanCondon: Thanks for reviewing @wenovus 

`Extra` is defined as a `map[string][]interface{}` and the `[]interface{}` had only 1 item ever added to it. Regardless of whether the item to be added was a single struct or an array of structs it was being added as 1 entry in the `[]interface{}`.

This created an unnecessary 2 step situation at time of decoding, where the item added was an array:
1) Open up the `[]interface{}` and discover it has only 1 entry
2) open up that 1 entry and iterate through its entries

What I am doing here, is when the item to be added is an array, is to add the entries individually to the `Extra` as I presumed it was meant to be used.

An alternative way to simplify it at a higher level would be to declare the `Extra` variable instead as `map[string]interface{}` and in this way the item could be set on it directly, regardless of whether single or array of struct.

I did not do this though as I thought changig the defintion of `Extra` might break existing dependencies.

pkg/yang/entry.go
SeanCondon: When running the tests it became clear that the existing implementation allowed a 2nd entry to overwrite the first. It was not duplicating, just overwriting.

wenovus: That's really curious. I'm realizing now that this `==` comparison is most likely comparing two slices, of a pointer value, which sticks out to me as something that may not be what's intended. How could the existing implementation, or now your loop implementation, overwrite when it's doing an append?

SeanCondon: @wenovus Thanks for replying so again so soon.
I found that the duplicate entry happened in `TestIfFeature()` when testing `LeafList`.
`LeafList` calls `ToEntry()` directly which calls `addExtraKeywordsToLeafEntry()` and then calls `addExtraKeywordsToLeafEntry()` again with the same "ifFeature" statment.

With the code on `master` when it populates `Extra` it adds every entry twice - then very cleverly the test case (`entry_test.go:2009`) only takes the first entry and discards the 2nd.

When I changed the way entries are written to the `Extra`, I changed the test case so it does not ignore any entries, but then I found that I was getting: 
```
entry_test.go:2146: leaf-list: did not get expected if-features, got [ft-leaf-list ft-leaf-list], want [ft-leaf-list]

```
because `LeafList` was adding the `Extra` twice.
That's why I added the statement to "Don't add again if identical" - yes it's comparing pointers - because that's what I wanted not to have the exact same thing added twice.

pkg/yang/entry.go
```diff
@@ -626,7 +626,6 @@ func ToEntry(n Node) (e *Entry) {
 			e.addError(err)
 		}
 		e.Prefix = getRootPrefix(e)
-		addExtraKeywordsToLeafEntry(n, e)
```
SeanCondon, line 629: not needed, since it was already being called in `ToEntry` above

pkg/yang/entry.go
SeanCondon: @wenovus both have been removed now and all tests are passing

wenovus (approved):
LGTM. @robshakir, does this look good to you?

SeanCondon:
> LGTM. @robshakir, does this look good to you?

Thanks @wenovus 
It would be great to get this merged if @robshakir is OK with it

PR #215, opened 2021-10-13
YangType.Equal should check whether inputs are nil

### LINKED ISSUE #178: YangType.Equal should check whether inputs are nil
Since it doesn't, it causes a nil pointer exception when inputs are nil.

wenovus:
Fixed #178

### DIFF
--- a/pkg/yang/yangtype_test.go
+++ b/pkg/yang/yangtype_test.go
@@ -1,3 +1,17 @@
+// Copyright 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


pkg/yang/yangtype_test.go
```diff
@@ -0,0 +1,53 @@
...
+			Name:           "foo",
+			Kind:           Ydecimal64,
+			FractionDigits: 5,
+		},
+		inRight: &YangType{
+			Name:           "bar",
+			Kind:           Ydecimal64,
+			FractionDigits: 5,
+		},
+		wantEqual: true,
```
robshakir, line 39: Test the unequal case where they're the same name but different types?

robshakir (approved):

pkg/yang/yangtype_test.go
wenovus: Done.

PR #216, opened 2021-10-21
Protect FindModuleBy functions with locks

kimmidi:
When used from multiple goroutines, the byNS and byPrefix maps can cause
concurrent write panics.


pkg/yang/modules.go
```diff
@@ -229,6 +236,9 @@ func (ms *Modules) FindModuleByNamespace(ns string) (*Module, error) {
 // FindModuleByPrefix either returns the Module specified by prefix or returns
 // an error.
 func (ms *Modules) FindModuleByPrefix(prefix string) (*Module, error) {
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
+
```
wenovus, line 241: This function (along with `byPrefix`) has been removed as of v0.3.0 (a backwards-incompatible version increment). Please use the function `FindModuleByPrefix` instead.

pkg/yang/modules.go
```diff
@@ -200,6 +204,9 @@ func (ms *Modules) FindModule(n Node) *Module {
 // FindModuleByNamespace either returns the Module specified by the namespace
 // or returns an error.
 func (ms *Modules) FindModuleByNamespace(ns string) (*Module, error) {
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
```
wenovus, line 208: Could you add a comment here that this is used to protect `byNS` from a race condition?

pkg/yang/modules.go
kimmidi: Sorry, did you mean to suggest a different FindModule* function ?

wenovus: Yes this method no longer exists in v0.3.0. There is a *function* of the same name in node.go

PR #219, opened 2022-01-04
Add unit tests for FindGrouping

wenovus:


pkg/yang/find_test.go
```diff
@@ -0,0 +1,290 @@
...
+					prefix d3;
+					namespace "urn:d3";
+
+					revision 01-01-01 { description "the start of time"; }
+
+					grouping gg { leaf b { type string; } }
+				}`,
+		},
+		inNode: func(ms *Modules) (Node, error) {
+			return FindNode(ms.Modules["dev2"], "g")
```
robshakir, line 246: If you change this test case to be:

```suggestion
			return FindNode(ms.Modules["dev"], "c")
```

then the test will fail. This is what @aashaikh was running into before and needed the fix referenced in https://github.com/openconfig/goyang/commit/5057f4840240f76faba4f87d84ff51e75597f224

wenovus: Is it correct to search for `gg` from `dev:c` though? It seems to me we shouldn't be demanding `FindGrouping`, which searches from a node, to try to find something from a module it doesn't import from.

I suggested a fix here: https://github.com/openconfig/oc-tools/pull/19#discussion_r778459045

robshakir: Good question!  I think this comes down to the semantics of the function, our current docstring reads:

```
// FindGrouping finds the grouping named name in one of the parent node's
// grouping fields, seen provides a list of the modules previously seen
// by FindGrouping during traversal.  If no parent has the named grouping,
// nil is returned. Imported and included modules are also checked.
```

From this I read that if I have a node `b` that has a parent `a` then I should be able to find a grouping defined in the scope of `a` using `b` as the context node. If `b` is a grouping itself and `a` is a module, then all imports could be argued to be in the context of `a`. I think I agree that if the module that defines `a` doesn't import the module that defines the grouping that is being referenced then that should fail (since one couldn't `use` that grouping in that context).

Let's clarify the docstring to say that the context node that is used for `FindGrouping` defines what the addressable elements are according to YANG namespace rules and then add a test that explicitly ensures that this fails. :-)

wenovus: Clarified docstring and added testcase :)

robshakir (approved):
Thanks for this! LGTM.

PR #221, opened 2022-03-03
No such module

johannemitzcisco:
Not sure how the project is dealing with yang 1.1 but I believe that the following need to be fixed.

### DIFF
--- a/pkg/yang/yang.go
+++ b/pkg/yang/yang.go
@@ -624,13 +624,13 @@ type Uses struct {
 	Parent     Node         `yang:"Parent,nomerge" json:"-"`
 	Extensions []*Statement `yang:"Ext" json:"-"`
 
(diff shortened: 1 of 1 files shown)


wenovus:
This was an attempt to do this before (https://github.com/openconfig/goyang/pull/94) but since this is a breaking change there was a discussion on how best to handle it: https://github.com/openconfig/goyang/issues/51#issuecomment-475042682

Given we have just released v1.0.0 it seems appropriate to not violate backwards compatibility and simply introduce the `Augments` field instead as was the latest intention in the link above. This field is not being currently used by the goyang library but evidently users are using it.

@robshakir unfortunately I didn't record any decisions we may have made on this when we went through all the pending issues in preparation for v1.0.0. Let us know if you have objections to introducing an `Augments` field to support YANG1.1.

Issue #75, opened 2018-03-30
"augment" inside "uses" doesn't work


idryzhov:
Examples of YANG module with "augment" inside "uses" are [ietf-ipv4-unicast-routing](https://tools.ietf.org/html/rfc8349#section-8) and [ietf-ipv6-unicast-routing](https://tools.ietf.org/html/rfc8349#section-9) from [RFC 8349](https://tools.ietf.org/html/rfc8349).

There is also a problem with multiple "augment" in one "uses" but it is already discussed in #51.

PR #222, opened 2022-03-03
Fix problem where a Choice statement contains Choice statements (yang…

johannemitzcisco:
Fix problem where a Choice statement contains Choice statements (yang1.1) and Issue https://github.com/openconfig/goyang/issues/75 (augments in Uses)

Signed-off-by: Johan Nemitz <johannemitzcisco@gmail.com>

### DIFF
--- a/pkg/yang/yang.go
+++ b/pkg/yang/yang.go
@@ -487,6 +487,7 @@ type Choice struct {
 	Anydata     []*AnyData   `yang:"anydata"`
 	Anyxml      []*AnyXML    `yang:"anyxml"`
 	Case        []*Case      `yang:"case"`
+	Choice      []*Choice    `yang:"choice"`
 	Config      *Value       `yang:"config"`
 	Container   []*Container `yang:"container"`
 	Default     *Value       `yang:"default"`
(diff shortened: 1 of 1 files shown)


wenovus:
Copying my comment from https://github.com/openconfig/goyang/pull/221:

There was an attempt to do this before (https://github.com/openconfig/goyang/pull/94) but since this is a breaking change there was a discussion on how best to handle it: https://github.com/openconfig/goyang/issues/51#issuecomment-475042682

Given we have just released v1.0.0 it seems appropriate to not violate backwards compatibility and simply introduce the `Augments` field instead as was the latest intention in the link above. This field is not being currently used by the goyang library but evidently users are using it.

@robshakir unfortunately I didn't record any decisions we may have made on this when we went through all the pending issues in preparation for v1.0.0. Let us know if you have objections to introducing an `Augments` field to support YANG1.1.

robshakir:
I'm OK with adding an `Augments` statement here, and gradually phasing out `Augment` if we think that this is something that would be reasonable to do. I suggest we populate both fields in the intermediate phase.

The `Choice` change looks backwards compatible.

Let's ensure that we have test coverage here before we merge it please :-)

wenovus:
I think Paul's comment here is what I'm supporting we implement: https://github.com/openconfig/goyang/issues/51#issuecomment-473492126

Johan let us know if you need any help with this.

PR #225, opened 2022-03-03
Unknown type

johannemitzcisco:
This fixes a type that is defined in another submodule.  This is in the Types.go file and the function:
func (d *typeDictionary) findExternal(n Node, prefix, name string)


johannemitzcisco:
This also fixes the issue when it is reported that there is a duplicate module reported even though the duplicates are in the same directory.  This is found in Modules.go in the function:
func (ms *Modules) add(n Node)

PR #229, opened 2022-05-25
enum values are not exported in rawSchema JSON Fixes #227

### LINKED ISSUE #227: enum values are not exported in rawSchema JSON
When a YANG file contains an enumeration the enumerated values are not output to the RawSchema because the `EnumType` ([link](https://github.com/openconfig/goyang/blob/de640d098cc1ef2b962c0fca3d3b710ebe1113ea/pkg/yang/types_builtin.go#L570)) is defined with fields with lower case initial letter (and hence internal).

For instance calling YGOT `generator` tool on `openconfig-system@2017-07-06.yang` will produce a Schema with out any enumeration values - (3rd last line below)

e.g. for severity
```json
"severity": {
    "Name": "severity",
    "Description": "Specifies that only messages of the given severity (or\ngreater severity) for the corresonding facility are logged",
    "Kind": 0,
    "Config": 0,
    "Prefix": {
        "Name": "oc-log",
        "Source": {
            "Keyword": "prefix",
            "HasArgument": true,
            "Argument": "oc-log"
        }
    },
    "Type": {
        "Name": "syslog-severity",
        "Kind": 14,
        "Enum": {}
    }
}
```

It would be useful to be able to get the list of values from the enum
The expected values would look like the following (it would suffice if either one of `ToString` or `ToInt` was made public):
```json
"Enum": {
    "Last": 7,
    "ToString": {
        "0": "EMERGENCY",
        "1": "ALERT",
        "2": "CRITICAL",
        "3": "ERROR",
        "4": "WARNING",
        "5": "NOTICE",
        "6": "INFORMATIONAL",
        "7": "DEBUG"
    },
    "ToInt": {
        "ALERT": 1,
        "CRITICAL": 2,
        "DEBUG": 7,
        "EMERGENCY": 0,
        "ERROR": 3,
        "INFORMATIONAL": 6,
        "NOTICE": 5,
        "WARNING": 4
    }
}
```

SeanCondon:
Making `ToInt` and `ToString` public on the `EnumType`

### DIFF
--- a/pkg/yang/types_builtin.go
+++ b/pkg/yang/types_builtin.go
@@ -568,12 +568,12 @@ func pow10(e uint8) uint64 {
 // A EnumType represents a mapping of strings to integers.  It is used both
 // for enumerations as well as bitfields.
 type EnumType struct {
-	last     int64 // maximum value assigned thus far
-	min      int64 // minimum value allowed
-	max      int64 // maximum value allowed
-	unique   bool  // numeric values must be unique (enums)
-	toString map[int64]string
-	toInt    map[string]int64
+	last     int64            // maximum value assigned thus far
+	min      int64            // minimum value allowed
+	max      int64            // maximum value allowed
+	unique   bool             // numeric values must be unique (enums)
+	ToString map[int64]string `json:",omitempty"`
+	ToInt    map[string]int64
 }
 
 // NewEnumType returns an initialized EnumType.
@@ -583,8 +583,8 @@ func NewEnumType() *EnumType {
 		min:      MinEnum,
 		max:      MaxEnum,
 		unique:   true,
-		toString: map[int64]string{},
-		toInt:    map[string]int64{},
+		ToString: map[int64]string{},
+		ToInt:    map[string]int64{},
 	}
 }
 
@@ -596,8 +596,8 @@ func NewBitfield() *EnumType {
 		last:     -1, // +1 will start at 0
 		min:      0,
 		max:      MaxBitfieldSize - 1,
-		toString: map[int64]string{},
-		toInt:    map[string]int64{},
+		ToString: map[int64]string{},
+		ToInt:    map[string]int64{},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


SeanCondon:
retest this please

robshakir:
LGTM, thanks for the contribution - please could you re-push to the branch, it seems that the CI didn't run to test this.

We should add a TODO to have something that tests the expected JSON output for these types if they are to become a public API. @wenovus and I will discuss this and we can figure out the right thing to do.

robshakir (approved):

SeanCondon:
Thanks Rob. @wenovus - are you OK with this change?

pkg/yang/types_builtin.go
```diff
@@ -568,12 +568,12 @@ func pow10(e uint8) uint64 {
...
-	max      int64 // maximum value allowed
-	unique   bool  // numeric values must be unique (enums)
-	toString map[int64]string
-	toInt    map[string]int64
+	last     int64            // maximum value assigned thus far
+	min      int64            // minimum value allowed
+	max      int64            // maximum value allowed
+	unique   bool             // numeric values must be unique (enums)
+	ToString map[int64]string `json:",omitempty"`
+	ToInt    map[string]int64
```
wenovus, line 576: Could you add comments for these now that they're public?

SeanCondon: added - I added `name` and `value` from the [RFC 6020 9.6.4](https://datatracker.ietf.org/doc/html/rfc6020#page-120:~:text=cannot%20be%20restricted.-,9.6.4,-.%20%20The%20enum%20Statement)

wenovus (approved):
Thanks this LGTM. @SeanCondon I'm wondering if you need a release now or would you be fine with waiting a bit until @robshakir and I agree on what the right JSON representation is.

I think the current JSON representation is OK. There is no chance for conflict, and the names & integers match what the YANG spec says they should be.

wenovus:
> @wenovus - re the "github action workflow isn't running" - I added a specific unit test for the Enum JSON serialization to push the patch again, but the tests are still not run. I think this action may have been broken for a while, because when I run `go test ./...` on the master branch, I can see there are some failures there in relation to Decimal arithmetic which are unrelated to this change

The checks look to be running now (9 successful checks), are you seeing/expecting something else on your end? I just ran `go test ./...` locally on master and it looks to be passing.

SeanCondon:
@wenovus - I think I had an environment problem - tests are passing for me now.
I can wait for the next release, so please go ahead and merge. Thanks for review

PR #230, opened 2022-06-07
Fix workflow syntax and run on every pull request

wenovus:


robshakir (changes requested):
Oops, thanks for fixing this.

Could you check why static analysis is failing? It might be a transient issue.

robshakir (approved):

PR #242, opened 2023-04-21
Add `OrderedByUser` field

wenovus:
Per https://datatracker.ietf.org/doc/html/rfc7950#section-7.7.7 the field is either ordered by the user or it is not.

### DIFF
--- a/pkg/yang/entry_test.go
+++ b/pkg/yang/entry_test.go
@@ -3964,7 +3964,7 @@ type customTestCases struct {
 	wantEntryCustomTest func(t *testing.T, e *Entry)
 }
 
-func TestMisc(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/yang/entry_test.go
```diff
@@ -3958,3 +3958,165 @@ func TestLess(t *testing.T) {
 		}
 	}
 }
+
+type customTestCases struct {
+	wantEntryPath       string
+	wantEntryCustomTest func(t *testing.T, e *Entry)
+}
+
+func TestMisc(t *testing.T) {
```
DanG100, line 3967: TestList? misc is not descriptive

DanG100 (approved):

PR #243, opened 2023-05-02
Retain Extra fields over transforming Uses and Augments to Entry 

bclasse:
Hi,

In a YANG model, _uses_ or _augments_ can be used to extend modules. It's possible to define under those nodes an _IfFeature_ attribute. But when turning those into an Entry (_yang.ToEntry_), _IfFeature_ field disappears. It should become an attribute located under _Extras_, but when using _Uses_, it simply isn't copied hence some information is lost.

This PR adds support for moving all the Extra fields defined in Uses or Merge to the augment element.


bclasse:
CLA problem, will reissue this one.

PR #244, opened 2023-05-02
Retain Extra fields over Uses,Augments

bclasse:
Hi,

In a YANG model, uses or augments can be used to extend modules. It's possible to define under those nodes an IfFeature attribute. But when turning those into an Entry (yang.ToEntry), IfFeature field disappears. It should become an attribute located under Extras, but when using Uses, it simply isn't copied hence some information is lost.

This PR adds support for moving all the Extra fields defined in Uses or Merge to the augment element.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -1485,6 +1485,13 @@ func (e *Entry) merge(prefix *Value, namespace *Value, oe *Entry) {
 		} else {
 			v.Parent = e
 			v.Exts = append(v.Exts, oe.Exts...)
+			for lk, lv := range oe.Extra {
+				if v.Extra[lk] == nil {
+					v.Extra[lk] = lv
+					continue
+				}
+				v.Extra[lk] = append(v.Extra[lk], oe.Extra[lk]...)
+			}
 			e.Dir[k] = v
 		}
 	}
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -1461,6 +1461,12 @@ func (e *Entry) dup() *Entry {
 			ne.Dir[k] = de
 		}
 	}
+
+	ne.Extra = make(map[string][]interface{})
+	for k, v := range e.Extra {
+		ne.Extra[k] = v
+	}
+
 	return &ne
 }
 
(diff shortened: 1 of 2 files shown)


pkg/yang/entry.go
```diff
@@ -1485,6 +1485,13 @@ func (e *Entry) merge(prefix *Value, namespace *Value, oe *Entry) {
 		} else {
 			v.Parent = e
 			v.Exts = append(v.Exts, oe.Exts...)
+			for lk, lv := range oe.Extra {
+				if v.Extra[lk] == nil {
+					v.Extra[lk] = lv
+					continue
+				}
+				v.Extra[lk] = append(v.Extra[lk], oe.Extra[lk]...)
```
wenovus, line 1493: (optional) Note that currently only augments in `uses` statements are going to be picked up. Raw augments won't be. I'd also suggest adding this statement outside of the for loop as well and fixing the relevant test.

pkg/yang/entry.go
```diff
@@ -1485,6 +1485,13 @@ func (e *Entry) merge(prefix *Value, namespace *Value, oe *Entry) {
 		} else {
 			v.Parent = e
 			v.Exts = append(v.Exts, oe.Exts...)
+			for lk, lv := range oe.Extra {
+				if v.Extra[lk] == nil {
+					v.Extra[lk] = lv
+					continue
+				}
```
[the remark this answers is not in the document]
bclasse: Done

pkg/yang/entry.go
bclasse: Can you please give an example where this would occur?

bclasse: For what I see, `augment` defined in `uses` are ignored, regardless of if-feature field - could we open a new issue, approve this PR, and treat that corner case separately?  What do you reckon @wenovus ?

wenovus (approved):

pkg/yang/entry_test.go
```diff
@@ -2269,6 +2276,15 @@ func TestIfFeature(t *testing.T) {
 			inIfFeatures:   ms.Modules["if-feature"].Uses[0].IfFeature,
 			wantIfFeatures: []string{"ft-uses"},
 		},
+		{
```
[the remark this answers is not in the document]
bclasse: Done

wenovus:
Thanks!

PR #252, opened 2023-10-23
Fix nil pointer dereference panic in `Find()` if module by prefix is not found

### LINKED ISSUE #251: Nil pointer dereference panic in `Find()` if module by prefix is not found
Panic stack trace of goyang source files: 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x15a7cc0]

goroutine 20 [running]:
panic({0x18493c0, 0x1fe79a0})
        /usr/local/go/src/runtime/panic.go:884 +0x213 fp=0xc003b1ecd8 sp=0xc003b1ec18 pc=0x1036193
runtime.panicmem(...)
        /usr/local/go/src/runtime/panic.go:260
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:841 +0x37d fp=0xc003b1ed38 sp=0xc003b1ecd8 pc=0x104db1d
github.com/openconfig/goyang/pkg/yang.(*Module).ParentNode(0x159ea5f?)
        github.com/openconfig/goyang/pkg/yang/yang.go:143 fp=0xc003b1ed40 sp=0xc003b1ed38 pc=0x15a7cc0
github.com/openconfig/goyang/pkg/yang.RootNode({0x1af0840?, 0x0?})
        github.com/openconfig/goyang/pkg/yang/node.go:147 +0x45 fp=0xc003b1ed68 sp=0xc003b1ed40 pc=0x159f0e5
github.com/openconfig/goyang/pkg/yang.module({0x1af0840?, 0x0?})
        github.com/openconfig/goyang/pkg/yang/node.go:159 +0x27 fp=0xc003b1ed98 sp=0xc003b1ed68 pc=0x159f167
github.com/openconfig/goyang/pkg/yang.(*Entry).Find(0xc00660cf00, {0xc001625f80?, 0xc003b1ef40?})
        github.com/openconfig/goyang/pkg/yang/entry.go:1325 +0x425 fp=0xc003b1ee80 sp=0xc003b1ed98 pc=0x1595c85
```

https://github.com/openconfig/goyang/blob/944052f4c47385554f2fb0f46ca4ecfbf446450c/pkg/yang/entry.go#L1325

* `Find()` calls `FindModuleByPrefix()` which returns nil of type `*yang.Module`.
* `module()` is called with `(*yang.Module)(nil)`.

https://github.com/openconfig/goyang/blob/944052f4c47385554f2fb0f46ca4ecfbf446450c/pkg/yang/node.go#L158-L160

https://github.com/openconfig/goyang/blob/944052f4c47385554f2fb0f46ca4ecfbf446450c/pkg/yang/node.go#L146-L147

https://github.com/openconfig/goyang/blob/944052f4c47385554f2fb0f46ca4ecfbf446450c/pkg/yang/yang.go#L142

* `module()` calls `RootNode()` which expects the `Node` interface type. `(*yang.Module)(nil)` is passed into `RootNode()`.
* `ParentNode()` is called with `(*yang.Module)(nil)`. This is not `n == nil` because the interface has a nil pointer value (same explanation as [this](https://go.dev/doc/faq#nil_error)). This results in a nil pointer dereference panic.
* `RootNode()` is vulnerable to panics due to a non-nil interface (e.g., when a nil `*yang.Module` is passed). Even if we properly nil check in `RootNode()` with reflection, we will again be vulnerable to a panic again with `Kind()` in node.go:160.

The root cause here is that `module()` is called with `(*yang.Module)(nil)`. We should nil check `FindModuleByPrefix()` before calling `module()` to prevent the panic.

sengleung:
Previously, if `FindModuleByPrefix()` returns nil in `Find()`, it would continue to call `module()` with nil, resulting in a nil pointer dereference panic when trying to call `ParentNode()` from `RootNode()`. The fix is to return an error if `FindModuleByPrefix()` returns nil.

Fixes #251

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -1322,9 +1322,15 @@ func (e *Entry) Find(name string) *Entry {
 			e = e.Parent
 		}
 		if prefix, _ := getPrefix(parts[0]); prefix != "" {
-			m := module(FindModuleByPrefix(contextNode, prefix))
-			if m == nil {
+			mod := FindModuleByPrefix(contextNode, prefix)
+			if mod == nil {
 				e.addError(fmt.Errorf("cannot find module giving prefix %q within context entry %q", prefix, e.Path()))
+				return nil
+			}
+			m := module(mod)
+			if m == nil {
+				e.addError(fmt.Errorf("cannot find which module %q belongs to within context entry %q",
+					mod.NName(), e.Path()))
 				return nil
 			}
 			if m != e.Node.(*Module) {


LeonGWang (commented):
LGTM, thanks for the fix.

wenovus (approved):

wenovus:
Thanks for the fix

PR #260, opened 2023-11-07
Add `ClearEntryCache()` to `Modules`

### LINKED ISSUE #259: Cannot free up memory used by `entryCache` in `Modules`
`entryCache` is instantiated by `Modules`

https://github.com/openconfig/goyang/blob/a34098e7727d89e939164b3e0c0e4bd40c0bab5b/pkg/yang/modules.go#L35-L37

and is used only by the `ToEntry` function.

https://github.com/openconfig/goyang/blob/a34098e7727d89e939164b3e0c0e4bd40c0bab5b/pkg/yang/entry.go#L562-L567

`entryCache` in `Modules` can take up a bit of amount of memory after processing all YANG modules. For our use case, after processing all YANG modules, calls to `ToEntry`, which accesses the `entryCache`, happen very infrequently, if at all. Therefore, it is ideal to clear the cache to free up memory.

I don't think this should not compromise on correctness. Even if `entryCache` is cleared, calling `ToEntry` will just recurse and cache the relevant nodes again.

Issue #248 also refers to this issue. Creating this issue as this relates only to `entryCache`.

sengleung:
This function clears the `entryCache` in `Modules` to free up memory after processing modules. `entryCache` is used by the `ToEntry` function to cache previously converted nodes.

Fixes #259

### DIFF
--- a/pkg/yang/modules.go
+++ b/pkg/yang/modules.go
@@ -442,3 +442,9 @@ func (ms *Modules) include(m *Module) error {
 	}
 	return nil
 }
+
+// ClearEntryCache clears the entryCache containing previously converted nodes
+// used by the ToEntry function.
+func (ms *Modules) ClearEntryCache() {
+	ms.entryCache = map[Node]*Entry{}
+}


pkg/yang/modules.go
```diff
@@ -442,3 +442,9 @@ func (ms *Modules) include(m *Module) error {
 	}
 	return nil
 }
+
+// ClearEntryCache clears the entryCache containing previously converted nodes
+// used by the ToEntry function.
+func (ms *Modules) ClearEntryCache() {
+	ms.entryCache = map[Node]*Entry{}
```
wenovus, line 449: Suggest to clear both caches, since `Process` can't be called multiple times anyways.

A fix that allows multiple calls to `Process` should tolerate both caches being cleared between calls.

wenovus (commented):
goyang currently doesn't support calling `Process()` twice, despite what existing code comments may suggest right now.

This change LGTM for fulfilling your use case.

pkg/yang/modules.go
LeonGWang: What's the other cache (besides `ms.entryCache` you are suggesting to clear?)

wenovus: `mergedSubmodule`, but actually on another look, I'm concerned `ToEntry()` might not work anymore if that is also cleared. So scratch what I said, this PR looks good as-is.

wenovus (approved):

PR #268, opened 2024-07-10
Fix an age-old bug with `FromFloat` handling of negatives.

robshakir:
```
 * (M) pkg/yang/types_builtin(_test)?.go
    - This change squashes a bug that has existed for a long time
      in `FromFloat`, but only shows up on arm64. I finally have an
      arm64 machine and so found the issue. It was originally reported
      in https://github.com/openconfig/ygot/issues/766. The issue is
      that uint64 of a negative float64 is undefined  -- and on arm64
      on darwin returns 0, which made some test cases fail only on this
      architecture.
```

### DIFF
--- a/pkg/yang/types_builtin.go
+++ b/pkg/yang/types_builtin.go
@@ -328,18 +328,17 @@ func FromFloat(f float64) Number {
 		}
 	}
 
-	// Per RFC7950/6020, fraction-digits must be at least 1.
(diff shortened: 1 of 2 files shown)


wenovus (approved):
Nice find, thanks.

Interesting to know that Go also has undefined-ish behaviour for non-sensical code: https://go.dev/issue/52943

PR #272, opened 2024-08-08
Uses Augmentation

steiler:
Allow for the augmentation of groupings.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -660,6 +660,16 @@ func ToEntry(n Node) (e *Entry) {
 		// when the group is used in multiple locations and the
 		// grouping has a leafref that references outside the group.
 		e = ToEntry(g).dup()
+
+		// process augments
+		if s.Augment != nil {
+			a := ToEntry(s.Augment)
+			a.Parent = e
+			a.Augments = append(a.Augments, e)
+
+			e.Find(a.Name).merge(nil, a.Namespace(), a)
+		}
+
 		addExtraKeywordsToLeafEntry(n, e)
 		return e
 	}
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -661,12 +661,18 @@ func ToEntry(n Node) (e *Entry) {
 		// grouping has a leafref that references outside the group.
 		e = ToEntry(g).dup()
 
+		switch determineYangVersion(g) {
+		case YangVersion10:
+			if len(s.Augment) > 1 {
+				return newError(s, "multiple augments not allowed in yang version 1.0: %s", s.Name)
+			}
+		}
+
 		// process augments
-		if s.Augment != nil {
-			a := ToEntry(s.Augment)
+		for _, x := range s.Augment {
+			a := ToEntry(x)
 			a.Parent = e
 			a.Augments = append(a.Augments, e)
-
 			e.Find(a.Name).merge(nil, a.Namespace(), a)
 		}
 
@@ -1053,6 +1059,32 @@ func ToEntry(n Node) (e *Entry) {
 	}
 
 	return e
+}
+
+type YangVersion string
+
+const (
+	YangVersion10 YangVersion = "1.0"
+	YangVersion11 YangVersion = "1.1"
+)
+
+func determineYangVersion(n Node) YangVersion {
+	p := n.ParentNode()
+	if p != nil {
+		return determineYangVersion(p)
+	}
+	var m *Module
+	var ok bool
+	if m, ok = n.(*Module); ok {
+		switch m.YangVersion.asString() {
+		case string(YangVersion10):
+			return YangVersion10
+		case string(YangVersion11):
+			return YangVersion11
+		}
+	}
+	// default to 1.0
+	return YangVersion10
 }
 
 // addExtraKeywordsToLeafEntry stores the values for unimplemented keywords in leaf entries.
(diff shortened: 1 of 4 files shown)


pkg/yang/entry.go
```diff
@@ -1045,6 +1061,32 @@ func ToEntry(n Node) (e *Entry) {
 	return e
 }
 
+type YangVersion string
```
robshakir, line 1064: Can we add a docstring comment please?

pkg/yang/yang.go
```diff
@@ -624,13 +624,13 @@ type Uses struct {
...
 	Extensions []*Statement `yang:"Ext" json:"-"`
 
-	Augment     *Augment  `yang:"augment" json:",omitempty"`
-	Description *Value    `yang:"description" json:",omitempty"`
-	IfFeature   []*Value  `yang:"if-feature" json:"-"`
-	Refine      []*Refine `yang:"refine" json:"-"`
-	Reference   *Value    `yang:"reference" json:"-"`
-	Status      *Value    `yang:"status" json:"-"`
-	When        *Value    `yang:"when" json:",omitempty"`
+	Augment     []*Augment `yang:"augment" json:",omitempty"`
```
robshakir, line 627: Should we create a different field here, since this means that this change is backwards-incompatible for existing API users. There was some discussion historically about making such changes for YANG 1.1. 

If we don't do this, then we need to mark this as goyang v2. I'd prefer to see whether we can group up other API changes that would be needed to fully support y1.1 into such a release.

pkg/yang/entry_test.go
```diff
@@ -3871,6 +3871,260 @@ func TestLeafEntry(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ms := NewModules()
+			var errs []error
+
+			for n, m := range tt.inModules {
+				if err := ms.Parse(m, n); err != nil {
+					errs = append(errs, err)
+
```
robshakir, line 4091:
```suggestion
```

pkg/yang/entry_test.go
```diff
@@ -3871,6 +3871,260 @@ func TestLeafEntry(t *testing.T) {
...
+			}
+
+			var m *Entry
+			m, errs = ms.GetModule("mod-a")
+			x := m
+			for _, p := range tt.pathExist {
+				for _, pe := range p {
+					y, ok := x.Dir[pe]
+					if !ok {
+						t.Fatalf("expected module %s to contain path %s", m.Name, strings.Join(p, "/"))
```
robshakir, line 4117: Can we add more debugging output here -- e.g., "expected module %s to contain path %s, could not find element %s in parent, got children: %v"

pkg/yang/entry_test.go
```diff
@@ -3871,6 +3871,260 @@ func TestLeafEntry(t *testing.T) {
...
+				errsStr.WriteString(e.Error())
+			}
+
+			for _, errStr := range tt.WantErrors {
+				strings.Contains(errsStr.String(), errStr)
+			}
+
+			var m *Entry
+			m, errs = ms.GetModule("mod-a")
+			x := m
```
robshakir, line 4112: Suggest to make these variable names a little less pithy -- `curElem` or something would be easier for a reader to understand.

pkg/yang/entry.go
```diff
@@ -660,6 +660,22 @@ func ToEntry(n Node) (e *Entry) {
 		// when the group is used in multiple locations and the
 		// grouping has a leafref that references outside the group.
 		e = ToEntry(g).dup()
+
+		switch determineYangVersion(g) {
```
robshakir, line 664: We haven't added this kind of error checking elsewhere -- I wonder whether we should try and include the parsed YANG version somewhere in the output entry. WDYT?

pkg/yang/entry.go
```diff
@@ -1045,6 +1061,32 @@ func ToEntry(n Node) (e *Entry) {
 	return e
 }
 
+type YangVersion string
+
+const (
+	YangVersion10 YangVersion = "1.0"
```
robshakir, line 1067: `s/YangVersion10/YANGVersion10/` per usual Go style guidelines.

a-chartier:
This would be fantastic to have merged in.

pkg/yang/entry.go
steiler: done

pkg/yang/entry_test.go
steiler: done

pkg/yang/entry_test.go
steiler: done

pkg/yang/entry.go
steiler: done

pkg/yang/yang.go
steiler: I'm unsure what exactly you want me to do here.

pkg/yang/entry.go
steiler: if we want to be strict then thats probably a good idea.

pkg/yang/yang.go
robshakir: I don't support changing `Augment` from `*Augment` to `[]*Augment` since this is backwards incompatible. This means that we're going to need to all downstream clients to change.

I suggest this is implemented using a new field `Augments` which has the slice. If it's not, then I don't think we should merge this without discussing other cardinality changes that are needed to support YANG 1.1 in a goyang v2.

PR #274, opened 2024-11-12
Add must keyword support to input statement

alexandernorth:
added must substatement under input type, conforming to yang-version 1.1 as in https://datatracker.ietf.org/doc/html/rfc7950#section-7.14.2.1


alexandernorth:
> duplicate of #270 ?

You are right, sorry. Will close in favour of your PR.

Issue #269, opened 2024-07-10
must doesn't seem to be supported in deviation


ankitjain093:
the following example of a deviation statement to add a must constraint to a leaf doesn't seem to be supported:

```
deviation /l2:l2vpn-svc/l2:sites/l2:site/l2:site-network-accesses/l2:site-network-access/l2:connection/l2:lag-interfaces/l2:lag-interface/l2:lacp/l2:mini-link-num {
        deviate add {
            must "count(../member-links/member-link) >= ." {
                error-message "Minimum links can't be less than configured member links";
            }
        }
    }
```

PR #278, opened 2025-02-13
Allow for the deviation of Must Statements

steiler:
This is a PR that allow for the deviation of Must Statements.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -1278,7 +1278,7 @@ func (e *Entry) ApplyDeviate(deviateOpts ...DeviateOpt) []error {
 								devNodeMust[idx] = devNodeMust[len(devNodeMust)-1]            // Swap with last element
 								deviatedNode.Extra["must"] = devNodeMust[:len(devNodeMust)-1] // Trim last element
 							} else {
-								appendErr(fmt.Errorf("must statement marked for deletion not found [%s]", existingMust.Name))
+								appendErr(fmt.Errorf("must statement marked for deletion not found [%s]", mustDelete.Name))
 							}
 						}
 					}
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -1190,6 +1190,10 @@ func (e *Entry) ApplyDeviate(deviateOpts ...DeviateOpt) []error {
 						deviatedNode.Type = devSpec.Type
 					}
 
+					if musts, exists := devSpec.Extra["must"]; exists {
+						deviatedNode.Extra["must"] = append(deviatedNode.Extra["must"], musts...)
+					}
+
 				case DeviationNotSupported:
 					dp := deviatedNode.Parent
 					if dp == nil {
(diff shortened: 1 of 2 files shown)


steiler:
should also help with #269

pkg/yang/entry.go
```diff
@@ -1248,6 +1252,37 @@ func (e *Entry) ApplyDeviate(deviateOpts ...DeviateOpt) []error {
...
+						for _, mustDeleteInterf := range musts {
+
+							mustDelete := mustDeleteInterf.(*Must)
+
+							found := false
+							var idx int
+							var existingMustInterf interface{}
+							var existingMust *Must
+							for idx, existingMustInterf = range deviatedNode.Extra["must"] {
+								if existingMustInterf.(*Must).Name == mustDelete.Name {
```
robshakir, line 1266: It might be worth pulling this logic out into a helper function that states what it is doing a bit more clearly -- for example, something like `isSameMustStmt`, given that I assume that we actually need to handle other kind of deviations for `must` statements, right? (did not check rfc7950 here)

pkg/yang/entry.go
```diff
@@ -1248,6 +1252,37 @@ func (e *Entry) ApplyDeviate(deviateOpts ...DeviateOpt) []error {
...
+									existingMust = existingMustInterf.(*Must)
+									if (existingMust.ErrorMessage == nil && mustDelete.ErrorMessage == nil) ||
+										(existingMust.ErrorMessage != nil && mustDelete.ErrorMessage != nil &&
+											existingMust.ErrorMessage.Name == mustDelete.ErrorMessage.Name) {
+										found = true
+										break
+									}
+								}
+							}
+							if found {
```
robshakir, line 1276: Suggest to switch this around to keep the error case the indented one per go-style -- i.e., `if !found { appendErr(); continue }`.

robshakir (changes requested):
Thanks for the contribution!

PTAL at the comments, and please can we add a test for this functionality.

pkg/yang/entry.go
steiler: done

pkg/yang/entry.go
steiler: Changed the implementation meanwhile.
My take in the RFC was that all elements must match, but after some discussions seems like only the name (the actual must expression) must match, the error code etc does not matter. Hence the code is a lot simpler.
Change pushed

pkg/yang/entry.go
```diff
@@ -1248,6 +1252,37 @@ func (e *Entry) ApplyDeviate(deviateOpts ...DeviateOpt) []error {
...
+					if musts, exists := devSpec.Extra["must"]; exists {
+						// range through the deviations to apply them
+						for _, mustDeleteInterf := range musts {
+
+							mustDelete := mustDeleteInterf.(*Must)
+
+							found := false
+							var idx int
+							var existingMustInterf interface{}
+							var existingMust *Must
```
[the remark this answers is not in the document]
steiler: done

PR #287, opened 2025-04-24
Implement processing of refine statements, choice shorthand fixes

alexandernorth:
This PR implements the logic to process parsed refine statements according to the Yang 1.1 RFC.
It also implements a fix for choice cases, where the case statement is not present (allowed as per RFC). It does this by inserting the case automatically, such that a choice always has a case and no shorthand is used. This is important as paths are defined as if there is always a case node in the tree, even if it is missing in the schema.

### DIFF
--- a/pkg/yang/entry.go
+++ b/pkg/yang/entry.go
@@ -660,6 +660,83 @@ func ToEntry(n Node) (e *Entry) {
 		// when the group is used in multiple locations and the
 		// grouping has a leafref that references outside the group.
 		e = ToEntry(g).dup()
+
+		for _, refine := range s.Refine {
+			refineTarget := e.Find(refine.Name)
+			if refineTarget == nil {
+				return newError(s, "target node to refine %s not found in path: %s", refine.Name, e.Path())
+			}
+			// apply refinement according to https://datatracker.ietf.org/doc/html/rfc7950#section-7.13.2
+			// as best as we can
+
+			//   o  A leaf or choice node may get a default value, or a new default
+			//      value if it already had one.
+			//
+			//   o  A leaf-list node may get a set of default values, or a new set of
+			//      default values if it already had defaults; i.e., the set of
+			//      refined default values replaces the defaults already given.
+			if refine.Default != nil {
+				// TODO: do it
+				_ = refine.Default
+			}
+
+			//
+			//   o  Any node may get a specialized "description" string.
+			if refine.Description != nil {
+				refineTarget.Description = refine.Description.asString()
+			}
+
+			//
+			//   o  Any node may get a different "config" statement.
+			if refine.Reference != nil {
+				refineTarget.Config, err = tristateValue(refine.Reference)
+				if err != nil {
+					return newError(n, "error determining TriState value of Config: %s", err)
+				}
+			}
+
+			//   o  A leaf, anydata, anyxml, or choice node may get a different
+			//      "mandatory" statement.
+			if refine.Mandatory != nil {
+				refineTarget.Mandatory, err = tristateValue(refine.Mandatory)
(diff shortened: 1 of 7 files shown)


pkg/yang/entry.go
```diff
@@ -11,6 +11,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//
+// Copyright 2025 Swisscom (Schweiz) AG
```
robshakir, line 15: This reads like this whole file is `(c) Swisscom`, can you please adjust this to follow the first line and make it "additions copyright"? Alternatively, we can make this "(c) the OpenConfig contributors" and create a CONTRIBUTORS file.

pkg/yang/entry.go
```diff
@@ -660,6 +676,115 @@ func ToEntry(n Node) (e *Entry) {
 		// when the group is used in multiple locations and the
 		// grouping has a leafref that references outside the group.
 		e = ToEntry(g).dup()
+
+		for _, refine := range s.Refine {
```
robshakir, line 680: Given that this is a relatively long block -- can we split it out into its own function that we can also test individually?

pkg/yang/entry.go
```diff
@@ -660,6 +676,115 @@ func ToEntry(n Node) (e *Entry) {
...
+			if refine.Reference != nil {
+				refineTarget.Config, err = tristateValue(refine.Reference)
+				if err != nil {
+					return newError(n, "error determining TriState value of Config: %s", err)
+				}
+			}
+
+			//   o  A leaf, anydata, anyxml, or choice node may get a different
+			//      "mandatory" statement.
+
```
robshakir, line 730: (nit) remove blank like to match other cases.

pkg/yang/entry.go
```diff
@@ -755,12 +880,100 @@ func ToEntry(n Node) (e *Entry) {
 			}
 		case "choice":
 			for _, a := range fv.Interface().([]*Choice) {
+				//   https://datatracker.ietf.org/doc/html/rfc7950#section-7.9.2
```
robshakir, line 883: Again, can we pull this logic out into a separate function that we can test?

pkg/yang/yang.go
```diff
@@ -11,6 +11,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//
+// Copyright 2025 Swisscom (Schweiz) AG
```
robshakir, line 15: Same comment here re: copyright statements as the other file. (No problem noting the copyright, but we should just make it compliant with usual styles.)

pkg/yang/yang.go
```diff
@@ -646,7 +650,7 @@ type Refine struct {
 	Parent     Node         `yang:"Parent,nomerge"`
 	Extensions []*Statement `yang:"Ext"`
 
-	Default     *Value   `yang:"default"`
+	Defaults    []*Value `yang:"default"`
```
robshakir, line 653: This is a backwards incompatible change, that needs to be handled by downstream consumers too. 

We have had some discussion before on this: https://github.com/openconfig/goyang/issues/51

The conclusion was that we would have a `Default` and `Defaults` field, and this would allow handling of these cases in a backwards compatible way. Can we do this here so that we don't break the existing users?

robshakir (changes requested):
Thanks for the contribution, and extending goyang's functionality -- apologies for the delay in getting to review this contribution.

Please take a look at the comments.

PR #291, opened 2025-07-19
Fix github actions workflow for CI.

robshakir:


dplore (approved):

dplore:
> Ran into this in [openconfig/kne#604](https://github.com/openconfig/kne/pull/604) should be good here now.

Thank you sir!

